Issue #1179, opened 2026-07-02
Extension refractor 


HeartlessVeteran2:
![medium](https://www.gstatic.com/codereviewagent/medium-priority.svg)

Using `.first()` on `extensionRepository.getInstalledExtensions()` collects only the first emission of the flow. If the flow emits an empty list initially while loading, or if extensions are installed/uninstalled later, the library cards will not dynamically update their icons. Consider combining the flow of extensions directly with the library flow, or observing updates reactively instead of fetching a single snapshot with `.first()`.

_Originally posted by @gemini-code-assist in https://github.com/Heartless-Veteran/Otaku-Reader/pull/1178#discussion_r3510319520_

PR #1223, opened 2026-08-02
Unblock the extension loader: two fatal bugs meant no sources ever loaded

HeartlessVeteran2:
## 📋 Description

Stage 2 of the source-system rebuild, and the one that should make the app usable again.

**Two independent bugs each stopped extensions dead.** Either alone was enough to leave Browse showing nothing but the local source — and neither printed anything explaining why. That combination is why this looked like "the extension system is broken" rather than a set of specific defects.

### Bug 1 — the lib-version gate was set too low

`LIB_VERSION_MAX` was `1.5`, so every Keiyoushi/Komikku extension built against **extensions-lib 1.6 or 1.7** was rejected with `Unsupported lib version` *before a line of its code ran*. This is a compatibility **gate**, not a capability check — rejection happens up front.

The evidence that `1.7` is the right ceiling is inside this repo. The host contract in `core:tachiyomi-compat` already implements **10 members** marked `@since komikku/extensions-lib 1.6` (the related-manga APIs on `CatalogueSource`) and **one at 1.7** (`Source.getPageList`). The loader was refusing extensions that use APIs the app already ships.

Widening is safe because extensions-lib revisions are **additive** — 1.6/1.7 added members without breaking the 1.4-era ones, so both link against the same contract.

> ⚠️ Raise this further **only** after adding the corresponding members to `core:tachiyomi-compat`. Admitting an extension built against a revision the host doesn't implement produces a `LinkageError`, and *which* one depends on what's missing:
> - a missing host **class** → `NoClassDefFoundError` while the extension's class is resolved, i.e. at instantiation, so the loader catches it and reports "no valid sources";
> - a missing **method** on a class we do ship → **`NoSuchMethodError` at the call site**. Since additive revisions are exactly this case, that's the failure to expect — and it's the more dangerous one, because it escapes the loader entirely and surfaces later as a crash mid-browse or mid-read;
> - an interface member the extension doesn't implement but the host invokes → `AbstractMethodError`, likewise at call time.
>
> So the gate isn't a nicety: it converts a deferred, hard-to-trace runtime crash into a clear up-front rejection. A test pins **1.8 as rejected** for that reason.

### Bug 2 — Injekt had no `Application` binding

`ConfigurableSource.getSourcePreferences()` resolves `Application` through Injekt, but `OtakuReaderApplication` only registered `NetworkHelper` and `Json`.

Injekt is a **service locator**, so a missing entry throws at the call site rather than failing to compile — nothing catches it until a real extension loads on a device. Many real extensions read preferences from a **constructor** or a **`baseUrl` getter**, so they threw during instantiation and the loader recorded them as having no valid sources.

```kotlin
Injekt.addSingletonFactory<Application> { this }
```

### And the reason this was invisible

`refreshSources()` discarded every non-`Success` load result via `filterIsInstance<Success>()` **with no logging at all**. A fully broken pipeline produced an empty source list and *nothing in logcat*.

Failures are now logged grouped by a structured reason, with untrusted extensions reported separately since those are user-recoverable.

## 🐛 Four version defects, all in the parser

Review rounds turned up three more bugs in the same code path. Every one presented identically — a supported extension refused with `Unsupported lib version` — which is why they'd gone unnoticed:

| `versionName` | old behaviour | cause |
|---|---|---|
| `1.6.x` / `1.7.x` | rejected | ceiling set to 1.5 |
| `1.7` | rejected **as too old** | `substringBeforeLast('.')` → `"1"` → 1.0 |
| `1.4.19.1` | rejected | → `"1.4.19"`, not a number → `null` |
| `1.40.0` | **wrongly accepted** | Double `1.40` == `1.4` |
| `1.10.0` | rejected **as too old** | Double `1.10` == `1.1`, below the 1.4 minimum |

That last row is the one that mattered most: extensions-lib runs 1.7 → 1.8 → 1.9 → **1.10**, so the decimal comparison carried a latent mass-rejection bug set to fire the moment upstream's minor version hit double digits.

Version numbers aren't decimals. There's now a `LibVersion(major, minor)` value type with proper `Comparable` ordering, tested directly rather than only through the loader — because four separate parser bugs all collapsed into one indistinguishable rejection message.

## 🔄 Type of Change
- [x] 🐛 Bug fix
- [x] ♻️ Refactoring

## 🧪 Testing

Verified locally — I installed an Android SDK in the working container, so this was compiled and run rather than reasoned about:

| Check | Result |
|---|---|
| `:app:assembleDebug` | ✅ |
| `:data:testDebugUnitTest` | ✅ |
| `:core:extension:testDebugUnitTest` | ✅ |
| `:core:tachiyomi-compat:testDebugUnitTest` | ✅ |
| `detekt` | ✅ |
| `ktlintCheck` | ✅ |

**New coverage:** `LibVersionTest` (parsing and ordering, including `1.40` vs `1.4` and `1.10` vs `1.9`); lib versions 1.6/1.7 accepted and 1.8 rejected, asserted on structured `Reason` rather than message text; `ConfigurableSourcePreferencesTest` pinning the Injekt binding, including a source that reads preferences during construction; `SkippedExtensionSummaryTest` asserting 50 same-reason failures collapse to one log line.

**Two tests failed on first run**, which is what made compiling locally worth the setup:
- `Injekt.addSingletonFactory` caches the first resolved instance for the life of the JVM, so registering a fresh mock in `@Before` left later tests asserting against a stale one. Moved to `@BeforeClass` with per-test `clearMocks`.
- The original log grouping keyed on the message, which interpolates `$pkgName` — so it degenerated to one line per extension, exactly the flood it was meant to prevent.

**Still unverified:** that a real extension loads end-to-end on a device. That needs hardware — add the Keiyoushi repo, install an extension, confirm it appears in Browse and opens a chapter.

## 📸 Screenshots
N/A — no UI changes.

## ✅ Checklist
- [x] Code follows style guidelines
- [x] Self-review completed
- [x] Comments added for complex code
- [x] Documentation updated
- [x] No new warnings
- [x] Tests pass

### Also in this PR

- **Unified the two preference entry points.** The legacy free-function `sourcePreferences()` now delegates to `getSourcePreferences()` rather than duplicating its body, so a source can't observe different preferences depending on which form it calls. The receiver and signature are unchanged — extensions link against that symbol.
- **Deleted 1,121 lines of dead code:** `TachiyomiExtensionLoader` (389), its private `ChildFirstPathClassLoader` copy (103), and the 629-line test that only exercised itself. Nothing in production referenced them.
- **Committed the v40 Room schema export.** Never checked in, though 35–37 and 39 are tracked. Needed as the migration baseline for v41/v42 in later stages. (38 is also absent and isn't reconstructible from here — flagging, not fixing.)

## 🔗 Related Issues

Part of the source-system rebuild. Supersedes the diagnosis in #1179.


HeartlessVeteran2:
Review round addressed in `b5d6897`. Summary, including the one suggestion I did **not** take:

**Taken — Sourcery, message-substring assertions.** Correct, and it undercut the very tests this PR adds. The acceptance tests asserted `!message.contains("Unsupported lib version")`, which is vacuous under rewording — change the message and the assertion still passes, so the 1.6/1.7 regression guards quietly stop guarding anything with no test going red. `ExtensionLoadResult.Error` now carries a `Reason` enum and the tests assert on that. Message text stays human-facing and free to reword. `Reason` defaults, so the ~8 existing construction sites are untouched.

**Taken — cubic P2, per-package log lines.** Also correct, and it defeated the point of the previous commit: loader messages interpolate `$pkgName`, so grouping on the truncated message gave each extension its own group. Rejecting 50 extensions at the version gate would have produced 50 warnings — precisely the flood the grouping existed to prevent. Now keyed on the structured `Reason`.

I pulled the grouping out into `summarizeSkippedExtensions()` returning the lines, because `android.util.Log` is a no-op stub under unit tests — behaviour verified only at the logging call site wouldn't have been verified at all. There's now a test asserting 50 same-reason failures collapse to one line.

**Taken — cubic P3, wrong failure mode.** Right, and worth correcting because that comment is the guidance for any future decision to raise the ceiling. Since 1.6/1.7 are additive *member* additions, the real failure is `NoSuchMethodError` at the call site, not `NoClassDefFoundError` at instantiation. That's the more dangerous case, since it escapes the loader entirely and surfaces later as a crash mid-browse rather than being caught and reported as "no valid sources". Missing classes and unimplemented interface members are documented alongside it with their respective timings.

**Not taken — Sourcery, "ensure `logSkippedExtensions` is invoked on all paths producing `ExtensionLoadResult`s."** The premise doesn't hold. The only other production consumer, `SourceRepositoryImpl.loadExtension()`, already surfaces failures — it returns `Result.failure(throwable ?: IllegalStateException(message))`, which propagates to the caller. `refreshSources()` was the single silent path, because it's the only one that processes results in bulk and discards them with `filterIsInstance<Success>()`. Moving the logging closer to result generation would put it inside `ExtensionLoader`, which currently has no Android logging dependency and is a pure result-returning component — I'd rather keep it that way than log from the producer.

Also backed out some Stage 3 scaffolding that a `git add -A` swept into `b5d6897` — a version-catalog entry and a new module unrelated to this PR. It would have slipped an alpha-versioned dependency past a review that wasn't looking for one. It'll land in its own PR.

Verified locally: `assembleDebug`, `:core:extension` / `:data` / `:core:tachiyomi-compat` unit tests, `detekt`, `ktlintCheck`.

---

HeartlessVeteran2:
CodeAnt round — one real bug fixed in `f4aa173`, two findings need no change.

### ✅ Fixed: lib-version parser misread 2- and 4-component versionNames

Good catch, and it's the same class of bug this PR exists to fix — a current extension rejected by the gate with a message implying it was genuinely unsupported.

`substringBeforeLast('.')` assumes exactly three components:

| `versionName` | parsed as | result |
|---|---|---|
| `1.4.19` | `1.4` → 1.4 | accepted ✅ |
| `1.7` | `1` → **1.0** | **rejected — below minimum** |
| `1.4.19.1` | `1.4.19` → **null** | **rejected — not a number** |

So `"1.7"` was read as 1.0 and refused as *too old*, and a four-component version failed to parse at all. Both surfaced as `Unsupported lib version`, indistinguishable from a real incompatibility.

Now takes the leading `major.minor` pair explicitly, which is what the lib version actually is. Tests cover all three shapes plus a single-component rejection so the parser can't drift into accepting bare majors.

Worth noting this parser came from upstream, which has the same three-component assumption. It holds for the conventional `<libVersion>.<extVersion>` naming, but nothing guarantees it.

### ❌ Not a defect: `ConfigurableSource.sourcePreferences()` receiver

The finding states the legacy helper had a `Source` receiver and that this PR changed it to `ConfigurableSource`, breaking the ABI for older extensions. The premise doesn't hold — `main` already declares:

```kotlin
fun ConfigurableSource.sourcePreferences(): SharedPreferences =
    Injekt.get<Application>().getSharedPreferences(preferenceKey(), Context.MODE_PRIVATE)
```

The receiver was `ConfigurableSource` before this PR and still is. Only the **body** changed, to delegate to `getSourcePreferences()`. The JVM signature is byte-identical, so no extension compiled against the old helper is affected. Preserving the ABI here was deliberate and is called out in the code comment.

### ✅ Already resolved: unused `:core:js-runtime` module

Correct at the reviewed commit (`b5d6897`), but that was Stage 3 scaffolding a `git add -A` had swept in by accident — an alpha-versioned dependency and an empty module, neither mentioned in this PR's description. It was backed out in `3734089`/`a18c25a`. The module lands in its own PR with the actual engine, where the dependency can be reviewed on its merits.

Verified locally: `:core:extension` unit tests, `detekt`, `ktlintCheck`.

---

HeartlessVeteran2:
Both findings valid — fixed in `c29de3b`. The version one turned out worse than reported.

### Lib versions are now integer pairs, not decimals

cubic flagged that `1.40.0` is admitted as `1.4`. Correct — but the decimal comparison is wrong in **both** directions, and the direction it didn't flag is the one that will actually bite:

| `versionName` | as Double | gate result |
|---|---|---|
| `1.40.0` | `1.4` | **admitted** — unsupported extension slips through, fails later inside class loading |
| `1.10.0` | `1.1` | **rejected as below minimum** — a *newer* extension refused for looking too old |

That second row isn't hypothetical. extensions-lib runs 1.7 → 1.8 → 1.9 → **1.10**, so this carries a latent mass-rejection bug that fires the moment upstream's minor version reaches double digits. And it would present exactly like the 1.5-ceiling bug this PR exists to fix: every current extension refused with `Unsupported lib version`, no other symptom.

Version numbers aren't decimals. Added a `LibVersion(major, minor)` value type with proper `Comparable` ordering; `LIB_VERSION_MIN`/`MAX` are now `LibVersion` constants (neither is referenced outside this module).

I tested `LibVersion` **directly** rather than only through the loader. Every version defect found on this PR — the 1.5 ceiling, the 2-component `"1.7"`, the 4-component `"1.4.19.1"`, and now double-digit minors — has been in the *parser* rather than the gate, and every one presented as the same single indistinguishable rejection. Testing the parse and the ordering separately makes the actual defect visible. Coverage: `1.40` sorting above `1.4`, `1.10` sorting above both `1.9` and the ceiling, negative/non-numeric rejection, and the supported window asserted against parsed values rather than restating the constants.

### `PACKAGE_NOT_FOUND` split out of `APK_NOT_FOUND`

Agreed. The package-name lookup path receives a package name, never a file path, so classifying its failure as a missing *file* described a lookup that never happened. Now a distinct reason, documented as such.

Verified locally: `assembleDebug`, `:core:extension` and `:data` unit tests, `detekt`, `ktlintCheck`.

---

PR #1224, opened 2026-08-02
Add core:js-runtime — a QuickJS source engine in a sidecar process

HeartlessVeteran2:
## 📋 Description

Stage 3 of the source rebuild: a **second `MangaSource` backend** where sources are plain `.js` files. No APK install, no signature verification, no classloader — and no compiled ABI to drift against upstream, which is the structural problem behind the failures Stage 2 fixed one instance of.

Nothing is wired into `SourceRepositoryImpl` yet; that's Stage 4. This PR is the engine and its contract.

## ⚠️ Correction: the original argument for this design was wrong

The first version of this PR argued that the engine **had** to run out-of-process because no Android QuickJS binding could interrupt a running script. I based that on Maven Central's search API, which returns only the alpha line for `quickjs-kt`, and concluded the newest release was `1.0.0-alpha13`.

That was wrong. `maven-metadata.xml` lists stable releases through **1.0.9**, and 1.0.9 exposes `installInterrupt`, `requestInterrupt`, `interruptEvaluation()` and `setEvaluationTimeoutMillis` — a real in-VM interrupt. This PR now depends on 1.0.9 and `QuickJsHost` sets a 20s evaluation timeout.

So the "only a process kill can stop a runaway script" premise is retired. **The sidecar stays, for a different and better reason:**

QuickJS is C. A native fault — inside the parser, the regex engine, or GC, reachable from malformed HTML long before anyone writes a hostile script — kills whatever process it happens in, and no JavaScript-level interrupt is polled there. In-process, that is the whole app dying and taking reading position and download-queue state with it. Out-of-process, it is one source reporting an error.

The in-VM timeout handles a misbehaving **script**; the process boundary handles a faulting **engine**. Two different failure modes, and 1.0.9 only closed the first. The 20s in-VM budget is deliberately shorter than the client's 30s process budget so the cheap remedy acts first — an interrupt leaves the engine usable, a kill forces a rebind and a replay of every registered source.

Secondary benefits that fall out of the same boundary: an `isolatedProcess` runs under a permission-less UID with no network and no access to app data, so an engine-level compromise lands somewhere that can do nothing; and the JS heap plus every parsed Jsoup document sits outside the main process's memory, which matters in an app already holding decoded page bitmaps.

## ⚙️ Design consequences that don't retrofit cheaply

**Results return through a `ParcelFileDescriptor` pipe, not a `String`.** Binder transactions cap around 1 MB, which page lists — and especially novel chapter HTML in Stage 7 — exceed routinely. A `String` return passes every test and then throws `TransactionTooLargeException` on real content. The sidecar writes on its own thread, because a pipe buffer is ~64 KB and the client can't drain until the call returns; writing inline is a deadlock, not a slow path.

**HTTP stays in the main process.** The sidecar calls back over a binder so every request keeps the app's shared `OkHttpClient` — certificate pinning, rate limiting, the cookie jar, WebView-synced cookies/UA. Equally, a compromised sidecar gets **no independent network path**. The extra hop is a binder round-trip against a request that's already network-latency bound.

**Failures travel in-band** as `JsCallResult` rather than as binder exceptions, so a broken script is an ordinary source failure `SourceHealthMonitor` already knows how to handle — not a crash crossing a process boundary.

## 🔒 Sandboxing

A QuickJS context starts with **no I/O of any kind**. Capability arrives only through the three globals installed explicitly:

- `Client` — HTTP, routed back to the main process
- `Document` — Jsoup, handed out as **integer handles** rather than object references, so a script never receives a Kotlin object graph to walk
- `SharedPreferences` — **scoped per source**, since these frequently hold site credentials

Also: call arguments are injected as **JSON literals rather than concatenated**, so a quote or backslash in a search query can't terminate the literal and change the expression being evaluated. And a **fresh context per call**, not per source — a script can't accumulate state, so one page of results can't poison the next.

### Outbound requests are validated per redirect hop

The bridge is an SSRF primitive by construction: a source picks the URL. Automatic redirect following is therefore **disabled**, and each hop is checked against the scheme and private-address policy *before* it is issued. Inspecting the final URL afterwards is not a defence — OkHttp would already have connected to the internal host, which is the entire behaviour the blocklist exists to prevent; suppressing the body at that point only hides the result, while timing still reveals whether the host is alive.

Taking redirects away from OkHttp also means taking on the two rules it applies silently, both of which are reimplemented and tested here: credentials (`Authorization`, `Proxy-Authorization`, `Cookie`) are stripped on an origin change, and 301/302/303 become a bodyless GET while 307/308 preserve method and body.

> **Follow-up, deliberately not in this PR:** an OkHttp **network** interceptor runs once per hop *including redirects*, so the address policy could be enforced there while OkHttp keeps its own battle-tested redirect semantics — deleting ~60 lines of security-critical code in favour of ~15. That refactor is queued for the Stage 9 security review rather than done mid-review here.

## 🔄 Type of Change
- [x] ✨ New feature

## 🧪 Testing

| Check | Result |
|---|---|
| `:app:assembleDebug` | ✅ |
| `:core:js-runtime:testDebugUnitTest` | ✅ |
| `detekt` | ✅ |
| `ktlintCheck` | ✅ |

Tests cover the **wire protocol** — the one place the two processes can silently disagree, since they compile against shared types but a source's JSON is untrusted input neither side controls. Coverage includes unknown-field tolerance (a source adding a field upstream must not break here while working everywhere else), absent-field defaults, explicit nulls, error-kind round-tripping, and bodies containing quotes and backslashes. Filter serialization and the redirect policy predicates are tested directly.

**The tests failed on first run** — all 11, with `Serializer for class 'JsCallResult' is not found`. The serialization compiler plugin wasn't applied to the new module. `@Serializable` classes compile fine without it and fail only at runtime, so nothing before the test run would have caught it.

**Deliberately not tested here:** the sidecar itself. It needs an Android service and the native engine, so it can't run in a JVM unit test. Its behaviour — including the adversarial `while (true) {}` case — is verified on device in Stage 8, and the kill path is called out explicitly in Stage 9's security review.

## 📸 Screenshots
N/A — no UI.

## ✅ Checklist
- [x] Code follows style guidelines
- [x] Self-review completed
- [x] Comments added for complex code
- [x] Documentation updated
- [x] No new warnings
- [x] Tests pass

### Note on the dependency

`quickjs-kt` is pinned at **1.0.9** — the only maintained Kotlin QuickJS binding with coroutine-native async bindings, which Mangayomi-format extensions need throughout. Risk is contained by keeping every reference to it inside `:core:js-runtime`, behind the app's own `MangaSource` interface; swapping engines touches one class. All four ABIs ship (arm64-v8a, armeabi-v7a, x86, x86_64). Rationale is recorded in `libs.versions.toml` next to the version.

## 🔗 Related Issues

Part of the source-system rebuild. Follows #1223.


HeartlessVeteran2:
Eight findings, six needing code changes — fixed in `44c3091`. **Two of them invalidated claims this PR's description made**, so I've corrected those too.

### 🚨 The process boundary wasn't a security boundary

`android:process=":jsengine"` gives a separate process under the **same UID**. The engine kept `INTERNET`, the app's private files, and every granted permission — so my claim that "a compromised sidecar has no independent network path" was **false**. It could have opened its own sockets and ignored the bridge entirely.

Added `android:isolatedProcess="true"`: a dedicated permission-less UID, no filesystem, no network of its own. Routing HTTP through the main process is now *enforced* rather than conventional. The two attributes are load-bearing for different reasons — without `isolatedProcess` containment is theatre; without the separate process a hung script takes the app with it.

### 🚨 The wall-clock budget didn't work

`withTimeout` cannot preempt a blocking binder transaction or a blocking pipe read. Cancellation would have been ignored, the kill would never have fired, and the caller would have hung — **precisely the failure mode that moving out-of-process was meant to solve, reproduced one layer up.** That one's squarely mine.

A separate watchdog now kills the engine when the budget expires. Killing the remote process is what unblocks the transaction; the pending call dies with it and surfaces as an ordinary failure. The kill isn't cleanup *after* the timeout — it **is** the timeout mechanism.

Related correction: for an isolated process, releasing the last binding is what destroys it. `killBackgroundProcesses` wouldn't have reached it.

### The rest

| # | Finding | Fix |
|---|---|---|
| 3 | HTTP bridge returned a `String` while permitting 32 MB bodies — reintroducing the binder limit the engine path avoids | Both directions now use the same pipe helper |
| 4 | **SSRF**: any HTTPS host allowed, so a source could probe LAN services or `169.254.169.254` | Loopback/private/link-local/unique-local refused — and re-checked against the **final** URL, since OkHttp follows redirects internally and the pre-flight check only saw the first hop |
| 5 | `peekBody` truncated silently while returning `ok=true`, so a source could parse half a document as whole | Reads one byte past the limit to detect overflow and fails; limit 32 MB → 8 MB |
| 6 | `fetchSearchManga` accepted a `FilterList` and discarded it — every filtered search silently ran unfiltered | Serialized, **active filters only** (sending untouched ones makes sources treat "not selected" as a real constraint) |
| 7 | Per-page headers dropped, so hosts requiring `Referer` would 403 | Exposed via `pageHeaders()` so the data survives to Stage 4, where the image pipeline and the baseUrl Referer fallback land together |

### On the Promise finding — correct, and I can't prove the fix here

An async IIFE evaluates to the Promise *object*, so results would have arrived as `[object Promise]` and failed decoding as a **source** protocol error — blaming the extension for an engine bug.

Switched to top-level `await` in module mode, which is what quickjs-kt's suspend `evaluate()` is built around. But I want to be straight about the limit: **this cannot be verified in a JVM test**, because the native engine is Android-only. So rather than assume, there's now an explicit runtime check that fails loudly and accurately if an unresolved Promise ever comes back. Real confirmation happens on device in Stage 8.

Finding 2 (`dirtyPreferences` never consumed) was already addressed in `81f5eae` — I'd spotted it independently and removed the accessor, documenting that `set` is call-scoped with persistence landing in Stage 4.

Verified locally: `assembleDebug`, `:core:js-runtime` unit tests, `detekt`, `ktlintCheck`.

---

PR #1240, opened 2026-08-08
Search AniList for a manga's media id, and score the results

HeartlessVeteran2:
Stage 5b slice 4a. #1239 gave the details screen AniList metadata, but only for manga the user **already tracks on AniList** — the media id came from the AniList `TrackEntry.remoteId`, and an untracked manga has none. `MatchAniListMediaUseCase` (#1237) has been able to pick the right entry since the day it landed; nothing has ever handed it any candidates to pick from. This is the half that finds them.

## Why the tracker's existing search couldn't be reused

`AniListTracker.search` already exists, and it's the wrong shape twice over:

- **It asks only for `title { romaji english }`.** The matcher scores against *every* name a work is known by, so dropping `native` and `synonyms` throws away exactly the evidence that resolves a Japanese-titled source against an English AniList entry.
- **It returns `TrackEntry`** — a *tracking* record. Using it here would mean fabricating a list entry for a manga the user doesn't track, which is precisely the case this exists to serve.

So the search lives on the metadata API instead. That endpoint is unauthenticated (it needs to work for users who never logged into AniList) and already inherits the rate-limit interceptor and certificate pinning from the shared Retrofit instance, so neither had to be restated.

## The cascade, and why it stops early

AniList's search is a text index, not a matcher. A source that names a work `"Boku no Hero Academia (Official Colored)"` returns an empty page — there is no entry by that name — and no amount of clever scoring rescues an empty candidate list. So when a search finds nothing, the next known title is tried.

It **stops at the first search that returns anything.** Merging results from every title would widen the candidate list at the cost of one request per title against a limit that can hold a response for up to 90 seconds (#1236). One extra request to rescue a failed search is worth it; three more to slightly widen an already-populated list is not.

Stopping early costs nothing in *match* quality, and that's worth being precise about: whichever search produced the candidates, they're scored against the **whole** title set, so a candidate found via a synonym is still judged on how well it matches the source title. There's a test that pins this — the source title finds nothing, the native title finds two candidates, and the sequel-vs-original distinction is still decided by the source title's season marker.

Only an **empty** result advances to the next title. A failure aborts the cascade: a dead request would be dead for the next title too, and turning one into four is what the rate-limit interceptor exists to prevent.

## Shape

`AniListSearchRepository` is its own domain interface rather than a method on `MangaMetadataRepository`. That one owns a cache — Room reads, Room writes, a TTL, and offline-first guarantees it documents in as many words. This one owns no state at all: every call is a round trip and there's nothing to observe. Putting a method with none of those guarantees behind the interface that makes them seemed like the wrong trade.

Placeholder handling is at the boundary, reusing `PlaceholderTitles` (#1238). Sources emit `"?"` and `"??"` for a missing localized title, and **two placeholders match each other perfectly** — an entry whose english title is `"?"` would otherwise score 1.0 against a source whose title is also `"?"`.

## Scope

Search and match only. Nothing is persisted, no UI changes.

Where a confirmed `anilistId` lives is a schema decision I'd rather make deliberately than in passing — it's a migration, the least reversible thing in this app — and the wrong-match picker needs it in the same breath. Both are the next slice; I've asked about the storage shape separately.

## Verification

`:app:assembleDebug`, `testDebugUnitTest`, `:domain:test`, `detekt`, `ktlintCheck` — all green. 17 new tests (9 on the cascade, 8 on the mapping and error handling). The errors-first GraphQL check was confirmed to fail two tests when removed; the cascade's stop-early and abort-on-failure rules each have a test that constructs the case where the rule actually bites.

---

## Summary by Sourcery

Implement AniList title search and resolution for untracked manga by querying AniList metadata, mapping results into media candidates, and orchestrating multi-title searches before scoring matches.

New Features:
- Add AniListSearchRepository domain interface and AniListSearchRepositoryImpl to perform AniList title searches for manga and return media candidates.
- Introduce ResolveAniListMediaUseCase to search AniList across multiple meaningful titles and select the best matching media using existing matching logic.

Enhancements:
- Extend AniList metadata GraphQL schema and API with a dedicated search query and response model tailored for title-based media lookup, including placeholder-aware mapping.
- Limit AniList search attempts and per-page results, skip placeholder or duplicate titles, and treat empty results as a normal outcome to respect rate limits and improve robustness.

Tests:
- Add unit tests for ResolveAniListMediaUseCase covering search cascade behavior, stopping conditions, failure handling, and placeholder title scenarios.
- Add unit tests for AniListSearchRepositoryImpl validating GraphQL error handling, candidate mapping, placeholder stripping, and transport error propagation.

---
## Summary by cubic
Adds AniList title search and candidate scoring to resolve a manga’s media id when it isn’t tracked yet. Searches the metadata API with source and alternative titles, stops at the first non-empty result, then scores to pick the best match.

- **New Features**
  - Added `AniListMetadataApi.searchMedia` GraphQL query (romaji/english/native/synonyms; per-page capped).
  - Implemented `AniListSearchRepository` and `AniListSearchRepositoryImpl` to map results to `AniListMediaCandidate`, strip placeholder titles, de-duplicate titles case-insensitively with locale-invariant `lowercase()`, and fail fast on GraphQL errors.
  - Added `ResolveAniListMediaUseCase` to try titles in order, stop at first non-empty result, abort on failure, limit to 4 attempts, and score with `MatchAniListMediaUseCase`.
  - Bound `AniListSearchRepository` in `RepositoryModule`. 20 tests cover cascade behavior, mapping, and error handling.

- **Bug Fixes**
  - Treat malformed AniList search responses (no `data` or no `Page`) as failures, not empty results; only `media: []` is a successful “no candidates” outcome. Updated interface docs and tests.

<sup>Written for commit 1fd2692673d5e96d903fc3dc0efded06d60d511d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/HeartlessVeteran2/Otaku-Reader/pull/1240?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/data/src/main/java/app/otakureader/data/di/RepositoryModule.kt
+++ b/data/src/main/java/app/otakureader/data/di/RepositoryModule.kt
@@ -66,6 +66,8 @@ import app.otakureader.domain.scheduler.ExtensionUpdateScheduler
 import app.otakureader.domain.scheduler.LibraryUpdateScheduler
 import app.otakureader.domain.scheduler.ReminderScheduler
 import app.otakureader.data.metadata.AniListMetadataRepository
+import app.otakureader.data.metadata.AniListSearchRepositoryImpl
+import app.otakureader.domain.repository.AniListSearchRepository
 import app.otakureader.domain.repository.MangaMetadataRepository
 import app.otakureader.domain.repository.UpdateErrorRepository
 import app.otakureader.domain.repository.UpdateRunSummaryRepository
@@ -200,6 +202,9 @@ abstract class RepositoryModule {
 
     @Binds
     abstract fun bindMangaMetadataRepository(impl: AniListMetadataRepository): MangaMetadataRepository
+
+    @Binds
+    abstract fun bindAniListSearchRepository(impl: AniListSearchRepositoryImpl): AniListSearchRepository
 
     companion object {
         @Provides
--- a/data/src/main/java/app/otakureader/data/metadata/AniListMetadataApi.kt
+++ b/data/src/main/java/app/otakureader/data/metadata/AniListMetadataApi.kt
@@ -19,6 +19,16 @@ import javax.inject.Inject
 interface AniListMetadataService {
     @POST("graphql")
     suspend fun query(@Body body: MetadataGraphQlQuery): MetadataResponse
+
+    /**
+     * Same endpoint, different response type.
+     *
+     * Retrofit picks the deserializer from the declared return type, so a search cannot reuse
+     * [query] without making `MetadataResponse` cover both shapes — every field nullable, and a
+     * missing `Media` indistinguishable from a missing `Page`.
+     */
+    @POST("graphql")
+    suspend fun search(@Body body: MetadataGraphQlQuery): SearchResponse
 }
 
 @kotlinx.serialization.Serializable
@@ -48,4 +58,30 @@ class AniListMetadataApi @Inject constructor(
                 variables = buildJsonObject { put("id", anilistId) },
             )
         )
+
+    /**
+     * Search AniList for manga whose title resembles [title].
+     *
+     * [perPage] is capped rather than taken on trust: AniList's own maximum is 50, and the matcher
+     * gains nothing from a long tail — it scores every candidate, so a hundred near-misses only
+     * costs time and bandwidth to reject.
+     */
+    suspend fun searchMedia(title: String, perPage: Int = DEFAULT_PER_PAGE): SearchResponse =
+        service.search(
+            MetadataGraphQlQuery(
+                query = SEARCH_QUERY,
+                variables = buildJsonObject {
+                    put("search", title)
+                    put("perPage", perPage.coerceIn(1, MAX_PER_PAGE))
+                },
+            )
+        )
+
+    companion object {
+        /** Enough for the right entry to be present; short enough that scoring stays cheap. */
+        const val DEFAULT_PER_PAGE = 10
+
+        /** AniList rejects anything above this. */
+        const val MAX_PER_PAGE = 50
+    }
 }
--- a/data/src/main/java/app/otakureader/data/metadata/AniListMetadataQuery.kt
+++ b/data/src/main/java/app/otakureader/data/metadata/AniListMetadataQuery.kt
@@ -115,3 +115,57 @@ data class MetadataTitle(
     val native: String? = null,
     val userPreferred: String? = null,
 )
+
+/**
+ * Title search, used to find which AniList media a source manga corresponds to.
+ *
+ * ### Why this is not `AniListTracker.search`
+ *
+ * The tracker already has a search, and it is the wrong shape twice over. It asks only for
+ * `title { romaji english }`, while the matcher scores against **every** name a work is known by —
+ * dropping `native` and `synonyms` throws away the evidence that resolves a Japanese-titled source
+ * against an English AniList entry. And it returns `TrackEntry`, a *tracking* record, so using it
+ * here would mean inventing a fake list entry for a manga the user does not track — which is the
+ * exact case this exists to serve.
+ *
+ * ### Only what the matcher reads
+ *
+ * No cover, no format, no dates. [app.otakureader.domain.model.AniListMediaCandidate] holds titles
+ * and nothing else, so anything more would be fetched, parsed, and dropped. The wrong-match picker
+ * will need a cover and a year to be usable by a human; that slice extends the query and the
+ * candidate model together, where there is a consumer to justify each field.
+ */
+internal const val SEARCH_QUERY = """
+    query (${'$'}search: String, ${'$'}perPage: Int) {
+      Page(perPage: ${'$'}perPage) {
+        media(search: ${'$'}search, type: MANGA) {
+          id
+          title { romaji english native }
+          synonyms
+        }
+      }
+    }
+"""
+
+@Serializable
+data class SearchResponse(
+    val data: SearchData? = null,
+    val errors: List<MetadataError> = emptyList(),
+)
+
+@Serializable
+data class SearchData(
+    @SerialName("Page") val page: SearchPage? = null,
+)
+
+@Serializable
+data class SearchPage(
+    val media: List<SearchMedia> = emptyList(),
+)
+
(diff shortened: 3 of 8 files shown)


HeartlessVeteran2:
Three findings across @sourcery-ai and @codeant-ai. One is a doc-only change (7e3905f); two are declining, with the evidence.

### `lowercase()` needs an explicit `Locale.ROOT` — declining, the premise is inverted

Kotlin's no-argument `String.lowercase()` **is** the invariant-locale mapping. That is the entire reason it was introduced in Kotlin 1.5 to replace `toLowerCase()`, which used the default locale. `Locale.ROOT` would be a no-op; `Locale.getDefault()` would *introduce* the Turkish dotless-i bug rather than avoid it.

I verified this rather than quoting the docs at you — throwaway test, default locale forced to `tr`:

```kotlin
Locale.setDefault(Locale.forLanguageTag("tr"))
assertEquals("title",  "TITLE".lowercase())                     // passes
assertEquals("tıtle",  "TITLE".lowercase(Locale.getDefault()))  // passes
```

Both assertions pass, so the no-arg form is already locale-stable and the locale-sensitive form is the one that breaks. The probe is deleted; 7e3905f leaves a comment at the call site so the next reviewer doesn't have to re-derive it.

### Redundant placeholder check between the use case and the repository — declining

They guard different things, and removing either loses something:

- **`ResolveAniListMediaUseCase`** filters so a placeholder cannot consume one of the four `MAX_SEARCH_ATTEMPTS`. Without it, a manga whose alternative titles are `["?", "??", "Real Title"]` burns two of its four attempts on strings that cannot match anything.
- **`AniListSearchRepositoryImpl`** guards because it is a public repository method, and the next caller is the wrong-match picker taking a **user-typed** string. Without it, an empty search box spends a request against a rate limit that can hold a response for 90 seconds.

The overlap is real but it isn't duplication of a rule — it's the same predicate answering two different questions at two boundaries. Removing the repository's would be the worse of the two, since it's the one that protects callers that don't exist yet.

### The resolver isn't invoked by any production flow — accurate, and intentional

Correct as an observation: nothing calls it yet, and the details screen still derives the AniList id only from a tracker entry. That's the scope line in the PR description, not an oversight.

The reason it's a separate PR is #1225, which drew 18 review findings largely because preference persistence was bolted into what should have been a wiring change. One reviewable idea per PR has been worth the extra round since. The blocker was that wiring this up needs a schema migration to store the confirmed id, and I didn't want to pick that in passing — a migration is the least reversible thing in this app.

That's now decided: a `manga_anilist_link` table (`mangaId` PK, `anilistId`, `userConfirmed`, `matchedAt`, `CASCADE` from `manga`), keeping the durable link separate from the TTL'd metadata cache. The follow-up slice wires the resolver into the metadata flow, falls back to the stored link when there's no AniList `TrackEntry`, and adds the wrong-match picker. Below the 0.7 confidence threshold nothing will render and the user gets a manual "Link to AniList" action — a wrong synopsis looks authoritative and the user has no reason to doubt it, so no metadata beats wrong metadata.

Worth flagging honestly: I'd hold this differently if the follow-up were speculative. Code that is unreachable *indefinitely* is the `deleteOlderThan` mistake from #1238 — dead, and carrying a claim about what it did. This is unreachable for one PR with the blocker resolved and the next slice already scoped.

---

HeartlessVeteran2:
@cubic-dev-ai — two of three are correct and fixed in 1fd2692. The third I'm declining, with a measurement.

### P2, missing `data`/`Page` reported as an empty search — valid, fixed

`response.data?.page?.media.orEmpty()` collapsed three outcomes into one: a genuine empty page, a null `Page`, and a null `data`. Only the first is an answer.

The consequence lands in the caller, and it's exactly the case its own KDoc argues against — `ResolveAniListMediaUseCase` advances on empty and aborts on failure because *"a dead request would be dead for the next title too, and turning one into four is what the rate-limit interceptor exists to prevent."* A malformed response **is** that dead request, and it was taking the advance path. Worse, `refreshMetadata` already draws this line for a missing `Media`; search was the inconsistent one, which is the tell I should have caught.

Now:

```kotlin
val page = response.data?.page
    ?: return Result.failure(AniListMetadataException("AniList returned no result page for \"$title\""))
Result.success(page.media.map { it.toCandidate() })
```

Three tests added; the two malformed-response cases were confirmed to fail against the old expression.

### P3, the interface contract is inaccurate for placeholders — valid, fixed

The KDoc said an empty list means *"AniList genuinely has nothing under that name"*. False on the short-circuit path, where nothing is ever sent. That's the pattern the repo's self-review checklist opens with — a comment asserting a property the code has on one path and not another — so thank you for catching it on a doc rather than in a bug.

Rewritten to define empty as "no candidates" without attributing it to AniList, to name both paths, and to state that a malformed response is a failure rather than an empty list.

### P3, `lowercase()` uses the default JVM locale — declining, the premise is factually wrong

This is the second time this has been filed on this PR (Sourcery raised it first), and cubic reviewed `7e3905f`, which already carries a comment at that exact line explaining it. So rather than restate: Kotlin's no-argument `String.lowercase()` is **not** the default-locale mapping. It is defined as the invariant-locale mapping, and replacing default-locale `toLowerCase()` is the reason it exists.

Measured, default locale forced to `tr`:

```kotlin
Locale.setDefault(Locale.forLanguageTag("tr"))
assertEquals("title", "TITLE".lowercase())                     // passes
assertEquals("tıtle", "TITLE".lowercase(Locale.getDefault()))  // passes
```

Both pass. The no-arg form already collapses `I`/`i` the way the dedup intends; `Locale.ROOT` would be a no-op and `Locale.getDefault()` is the form that actually produces the dotless-ı you describe. The specific failure in the finding — "a source title and a synonym differing only in capital-I will not be treated as duplicates" — does not occur.

If it's useful signal: this one has now cost two review rounds across two bots on a line that was already commented. Happy to be shown wrong by a counter-example, but it would need to be a run, not a reading.

---

PR #1241, opened 2026-08-08
Auto-match untracked manga to AniList and store the link

HeartlessVeteran2:
## **User description**
Stage 5b slice 4b. #1240 gave the app a search-and-score path and nothing called it — which cubic correctly flagged as unreachable code. This wires it in, so a manga that **isn't** tracked on AniList gets metadata too, closing the gap #1239 shipped with.

## Schema v42: `manga_anilist_link`

A second table rather than a column on `manga_metadata`, even though that row already carries an `anilistId` and reusing it would need no migration at all.

The two have different lifetimes. `manga_metadata` is a **cache**: seven-day TTL, rewritten by every refresh, and `clearMetadata` exists to throw it away. A link the user picked by hand **cannot** be. Storing a correction in a row designed to be discarded means the correction is discarded with it — which defeats the picker that produced it.

Not a column on `manga` either: that entity is 38 columns of the library's core record, read by nearly every query in the app, and a third party's identifier doesn't belong in it.

```kotlin
@Entity(tableName = "manga_anilist_link", foreignKeys = [/* CASCADE from manga */])
data class MangaAniListLinkEntity(
    @PrimaryKey val mangaId: Long,
    val anilistId: Long,
    val userConfirmed: Boolean = false,
    val matchedAt: Long = 0L,
)
```

`userConfirmed` is the load-bearing field. Auto-matching runs whenever a manga has no link, so without a flag saying *a human chose this*, the next auto-match would overwrite a correction with the same wrong guess that made the correction necessary. The no-overwrite rule lives in `AniListLinkRepository.saveAutoLink` rather than at the call site — a rule enforced at each caller is one a future caller forgets.

Writes are serialized per manga with the same striped-mutex pattern as `AniListMetadataRepository`. `saveAutoLink` is read-decide-write, and the losing interleaving is the one that matters: an auto-match reads "no user link", is descheduled while the user picks the right entry, then writes its guess over the correction. The lock spans the check *and* the write.

## Only a confident match is persisted

Below `ACCEPT_THRESHOLD` the matcher is saying it couldn't separate a work from its sequel. Writing that attaches a wrong synopsis, wrong tags and a wrong score — all of which look authoritative and give the user no reason to doubt them. So nothing is stored and nothing renders. `AniListMatch.confident` has existed since #1237 for exactly this call and now has a caller.

## Resolution order is load-bearing, not a preference

`State.anilistMediaId` takes the AniList `TrackEntry.remoteId` first and the stored link second. That isn't just "the tracker is more trustworthy": `State.metadata` hides cached metadata whose `anilistId` disagrees with this value (the guard added in #1239), so if a stale stored link could shadow a live tracker entry, **correcting the tracker link would stop invalidating the cache**. There's a test pinning the order for that reason.

## Two guards, both verified by breaking them

- **The migration's schema validation** catches `userConfirmed` declared `TEXT` rather than `INTEGER`. Confirmed by doing it: with the wrong type the other three migration tests still pass and only `validatesAgainstTheExportedSchema` fails — which is exactly the bug class that fails on upgrade and never on a fresh install, since a fresh install builds the table from the entity.
- **The stored link has to drive the metadata fetch**, not just populate state. My first draft updated `anilistLink` and stopped, so an untracked manga would resolve its id and then never fetch anything with it — the entire point of the slice, missing. The test caught it; the fix was confirmed to fail again with the trigger removed.

Auto-matching is attempted once per screen visit. A search that legitimately finds nothing leaves the link null, so an unguarded trigger would search again on every re-emission of the link flow — one AniList request per emission for any manga AniList has never heard of, against a limit that can hold a response for 90 seconds.

## Scope

No picker yet. That's the next slice, and it needs `SEARCH_QUERY` and `AniListMediaCandidate` extended with cover, format and start year so a human can actually tell candidates apart — query and model extended together, with the picker as the consumer justifying each field.

## Verification

`:app:assembleDebug`, `testDebugUnitTest`, `:domain:test`, `detekt`, `ktlintCheck` — all green. 9 new `DetailsViewModelTest` cases and 4 new migration tests. `CLAUDE.md` updated for schema v42 and the new resolution order.

---

## Summary by Sourcery

Persist AniList links for manga and wire in auto-matching so untracked titles can receive AniList metadata, backed by a new durable link table and repository.

New Features:
- Introduce a persistent AniList link model and repository to record which AniList media each manga maps to, including whether the user confirmed it.
- Enable automatic AniList matching in the details screen for manga not yet linked or tracked, using AniList search and feeding synonyms as extra search terms.

Enhancements:
- Make the AniList media id resolution prefer the tracker entry and fall back to the stored link, ensuring cache invalidation behaves correctly.
- Guard auto-matching to run at most once per details screen visit and only persist confident matches, avoiding repeated or low-quality searches.

Build:
- Add Room schema v42 with the `manga_anilist_link` table and a migration from v41, plus DI wiring for the new DAO and repository.

Documentation:
- Update CLAUDE.md for schema v42 and to describe the new AniList link storage and metadata resolution behaviour.

Tests:
- Extend details view model tests to cover AniList auto-matching behaviour, resolution order, and metadata refresh driven by stored links.
- Add migration tests for v41→v42 to verify creation, cascading behaviour, and schema validation of the `manga_anilist_link` table.

___

## **CodeAnt-AI Description**
Auto-link untracked manga to AniList and load their metadata

### What Changed
- Untracked manga now get a single AniList search when their details open, using the title and cached alternative titles.
- Only confident matches are saved; uncertain guesses are discarded so incorrect metadata is not shown.
- Stored AniList links now provide metadata for manga without an AniList tracker entry, while an existing tracker entry remains authoritative.
- User-confirmed links are preserved, and links remain available after metadata cache refreshes or clears.
- Added database migration and coverage for durable links, upgrade validation, and cleanup when manga is deleted.

### Impact
`✅ AniList metadata for untracked manga`
`✅ Fewer incorrect manga matches`
`✅ User-selected AniList links survive cache clears`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>

---
## Summary by cubic
Automatically match untracked manga to AniList and save a durable link, so they get rich metadata on first load. Matching is gated on state readiness, runs once per visit, and never overwrites a user pick.

- **New Features**
  - Auto-match when no tracker `remoteId` or stored link; save only confident matches and immediately fetch metadata.
  - Resolution order: tracker `remoteId` > stored link to keep cache invalidation correct.
  - Preserve user picks with `userConfirmed`; serialize writes per manga; run matching once per screen visit.

- **Bug Fixes**
  - Trigger from state readiness (manga, tracker entries, stored link, cached metadata) so emission order can’t skip matching.
  - Start on metadata’s first emission; fallback searches use synonyms when available and don’t re-search when synonyms arrive later.
  - Added tests for readiness gating and the repository’s no-overwrite race; fixed KDoc placement and clarified a test name to reflect the once-per-visit guard.

<sup>Written for commit 1fccc94289a75eccf9eefcad318719e9d0ce685f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/HeartlessVeteran2/Otaku-Reader/pull/1241?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


HeartlessVeteran2:
@codeant-ai and @cubic-dev-ai independently reported the same two P2s, plus two P3s from cubic. All four valid, all fixed in bf8cefc.

### P2 — matching silently skipped when the manga loads after the link flow

Correct, and the reasoning in both reports is exactly right. `matchAniListMedia()` was called only from `observeAniListLink`'s collector and bailed when `_state.value.manga` was still null. Both are Room flows launched from `init`, launch order doesn't fix emission order, and the link flow has no reason to re-emit — so the one attempt per visit was dropped with nothing to explain why.

This is the **same mistake as the stale-metadata bug cubic's colleague caught on #1239**, one PR later: a condition that depends on several independently-collected values, checked inside whichever collector happened to fire. I wrote a long comment there about why deriving beats filtering-in-a-collector, and then did the filtering-in-a-collector thing again.

So the fix is the same shape. `State.isReadyToMatchAniList` is derived from the manga, the tracker entries, the stored link and the cached metadata, and the trigger collects `state` — whichever input completes last fires it.

### P2 — the synonym fallback skipped when metadata arrives late

Also correct, and it's the reason the readiness gate waits for *four* things rather than two. Fixing only the manga race would have left this one.

Waiting needed explicit `hasLoadedTrackEntries` / `hasLoadedLink` / `hasLoadedMetadata` flags, because `null` and `emptyList()` are ambiguous — they mean both "not loaded" and "loaded, and there's nothing". Without them, a manga that *is* tracked on AniList could be auto-matched purely because the tracker flow hadn't emitted yet. There's a test for that specific case. The wait costs nothing: every input is a Room flow, so each is guaranteed to emit once up front.

### P3 — `autoMatch_neverStoresAGuess` would pass if matching never ran

Right, and it's the checklist item this repo keeps relearning: a test whose name claims a rule has to construct the case where the rule bites. Asserting only the absence of a save is satisfied by a broken invocation too. Added `coVerify(exactly = 1) { resolveAniListMedia(any(), any()) }` so it proves the guess was consumed and discarded.

### P3 — no repository-level tests for the no-overwrite rule

Fair, and the sharpest of the four. The ViewModel tests stub `AniListLinkRepository`, so they prove the ViewModel *calls* `saveAutoLink` — not that `saveAutoLink` protects anything. Seven tests added against a fake DAO, including the interleaving the lock exists for.

**Worth reporting: that race test was wrong before it was right.** My first fake DAO delayed *before* reading, so a slow read returned post-interleaving state — and the test passed with the lock deleted. Fixed to snapshot-then-delay, which models the actual hazard (a read observes some state; by the time the caller acts on what it saw, the world has moved on). It fails without the lock now, and didn't before. A green test over an unlocked implementation is precisely the "passing test that asserts the wrong thing" failure, and I'd have shipped it if I hadn't run the removal check.

All three production guards verified by breaking them: removing the loaded-flag gating fails `autoMatch_doesNotStartBeforeEveryInputHasLoaded`; moving the trigger back into the link collector fails `autoMatch_stillRunsWhenTheLinkFlowEmitsBeforeTheMangaLoads`; deleting the mutex fails the race test.

---

Separately, on the earlier "no path to `saveUserLink` or `clearLink`" note: accurate, and the picker is the next slice. One correction to *"potentially preserving a wrong auto-match indefinitely"* — there is an interim escape, because `anilistMediaId` prefers the tracker entry over the stored link, so linking the manga properly on AniList overrides a bad auto-match. Adding a bare "unlink" action wouldn't help: clearing the link just lets auto-matching re-derive the same wrong guess on the next visit. Only a `userConfirmed` pick breaks that loop, which is why the picker is the fix rather than a smaller stopgap.

---

HeartlessVeteran2:
Both correct, fixed in 1fccc94.

**Stranded KDoc** — the readiness refactor inserted `observeAniListMatchReadiness` ahead of `matchAniListMedia` and left the latter's doc block above the new function, so the file had two consecutive `/** */` blocks and the function the first describes had none. Moved back down.

**The test name claimed the wrong rule**, and this one is worth more than a rename. `autoMatch_waitsForCachedSynonymsBeforeSearching` asserts close to the opposite of its name: matching runs on the first (null) metadata emission *with no synonyms*, and then doesn't re-search when synonyms arrive later. That's the once-per-visit guard.

The name was papering over a distinction I'd been sloppy about in my own commit message. **The gate waits for the metadata flow's first *emission*, not for synonyms to exist** — only the first is something code can wait on. In production the two coincide, because Room's first emission carries the cached row if there is one, so synonyms are present by the time matching starts. But they're not the same claim, and the test happens to model exactly the case where they diverge: a row appearing *later*, which Room wouldn't do for a row that already exists.

So it's renamed to `autoMatch_startsOnceMetadataHasEmittedAndDoesNotResearchLater`, and the comment now names the case it covers and points at `autoMatch_feedsCachedSynonymsBackInAsSearchTerms` for the positive one — synonyms present at first emission, and actually passed to the resolver.

That's the third finding on this PR in the "a test or comment asserts something the code doesn't quite do" family, which is the repo's most common defect class by a wide margin. Useful to have it caught in a name rather than in behaviour.

---

PR #1243, opened 2026-08-08
Add the wrong-match picker for AniList links

HeartlessVeteran2:
## **User description**
Stage 5b slice 4c. #1241 made auto-matching refuse to guess: below `ACCEPT_THRESHOLD` nothing is stored and nothing renders, because a wrong synopsis and wrong tags look authoritative and give the user no reason to doubt them. That's the right default — and it left the user with no metadata and **no recourse**. This is the recourse, and it gives `saveUserLink` its caller.

## The query and the candidate model grew together

`SEARCH_QUERY` deliberately fetched titles and nothing else in #1240, because `AniListMediaCandidate` held titles and nothing else and anything more would have been fetched, parsed and dropped. The picker is the consumer that justifies more: **a list reading "Kaguya-sama wa Kokurasetai" three times over is not a choice anyone can make.** `coverImage`, `format` and the start year are what separate a work from its sequel, its colored edition and its spin-off.

The matcher ignores all three, deliberately — a cover can't make a title match better.

## Behaviour, and why

- **Opens seeded with the manga's title and searches immediately.** The common case is that the source title was close enough and the matcher just wouldn't commit, so that case costs one tap. The query stays editable because the case that *actually* needs fixing is the one where the source's title isn't what AniList calls the work.
- **A pick calls `saveUserLink`** — `userConfirmed`, which auto-matching refuses to overwrite — **then forces a refetch.** `force = true` matters: the cached row still describes the old media and is inside its seven-day TTL, so an unforced refresh would decline to do anything at all.
- **Retyping cancels the in-flight search** rather than queueing behind it. A slow earlier result landing afterwards would show results for a query that's no longer in the box.
- **A search failure is shown**, unlike auto-matching's deliberate silence. The user asked for this and is waiting on it, so silence would read as "no results" — a different answer with a different next step.
- The overflow entry is always offered, not only when unmatched: it has to exist both when auto-matching declined to guess (nothing rendered, no clue why) and when it guessed wrong (something rendered, and it's wrong).

## Verified rather than assumed

The plan said `State.metadata`'s existing guard *should* hide the old media's metadata the moment the link changes, and to test rather than assume. It does, and there's now a test instead of an assumption — worth having, because the stale row is inside its TTL and survives a failed refetch by design, so that guard is the only thing between a corrected link and the previous entry's synopsis, tags and score.

Both new guards were confirmed by deleting them: dropping `force = true` and dropping the search cancellation each fail a test.

## One thing I'm flagging rather than papering over

`clearLink` and `clearMetadata` still have no production caller — only their own tests. That's the `deleteOlderThan` shape from #1238, and I don't want it to slide a third time.

I considered adding an "Unlink from AniList" action here and decided against it, because **it wouldn't do what its name implies**. Clearing the link just lets auto-matching re-derive the same guess on the next visit — a loop, not a removal. Genuinely removing AniList data for a manga needs a "never match this one" sentinel (a `userConfirmed` row meaning *no*), which has real design questions: what `anilistMediaId` returns, and how `requestMetadataRefresh` avoids firing on it.

So: do you want that sentinel? If yes it's a small follow-up slice. If no, I'd rather delete `clearLink`/`clearMetadata` from the interfaces than leave two methods whose KDoc describes behaviour nothing exercises.

## Verification

`:app:assembleDebug`, `testDebugUnitTest`, `:domain:test`, `detekt`, `ktlintCheck` — all green. 6 new `DetailsViewModelTest` cases. `CLAUDE.md` updated with the picker as the documented recourse.

---

## Summary by Sourcery

Add an AniList wrong-match picker to the manga details screen, allowing users to search AniList manually, pick the correct entry, and update metadata safely when auto-matching is unsure or wrong.

New Features:
- Introduce an AniList picker dialog from the details overflow menu that lets users search for and select the correct AniList entry for a manga.
- Seed AniList searches with the manga title and allow editable queries, with results showing cover, format, and year to distinguish similar entries.
- Record a selected AniList candidate as a user-confirmed link and trigger a forced metadata refresh for the new AniList id.

Enhancements:
- Extend AniList search to return cover image, format, and start year and propagate these into the `AniListMediaCandidate` domain model for UI consumption.
- Expose a derived `displayTitle` and formatted subtitle for AniList candidates to present clearer choices in the picker.
- Ensure details state hides stale AniList metadata as soon as the link changes while retaining cached metadata separately.

Documentation:
- Update `CLAUDE.md` to document the AniList wrong-match picker as the user-facing recourse when auto-matching does not persist a match.

Tests:
- Add DetailsViewModel test coverage around opening the AniList picker, search behaviour, failure reporting, blank queries, link correction, and forced metadata refresh.

---
## Summary by cubic
Adds an AniList wrong‑match picker to the Details screen so users can link the correct entry when auto‑matching won’t commit or guessed wrong. Picks are saved as confirmed and trigger an immediate, forced metadata refresh; search is submit‑driven and clearly reports failures.

- **New Features**
  - Overflow → “Link to AniList…” opens `AniListPickerDialog` seeded with the manga title and runs a search.
  - Results show a display title, cover, `format`, and start year; selecting saves via `saveUserLink` as `userConfirmed`, then calls `refreshMetadata(..., force = true)`.
  - Submit‑driven search: only submitting replaces an in‑flight search; editing the query alone doesn’t cancel.

- **Bug Fixes**
  - Failed searches show a generic error; exception text is never rendered, and `searchFailed` keeps failures distinct from “no results”.
  - Cancellations propagate and aren’t treated as failures, so dismissing or resubmitting never flashes an error.
  - Untitled entries render “(Untitled)” and missing covers show a placeholder icon.
  - Blank queries clear results and skip the request; closing the picker cancels any in‑flight search.

<sup>Written for commit 37b750aef695047b22989ef27323c96d24f10804. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/HeartlessVeteran2/Otaku-Reader/pull/1243?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

___

## **CodeAnt-AI Description**
Let users choose the correct AniList entry when automatic matching is uncertain or wrong

### What Changed
- Added a “Link to AniList…” option that opens a searchable picker for any manga
- Searches start with the manga title and can be edited; results show covers, format, and publication year to distinguish similar entries
- Selecting an entry saves it as a user-confirmed link and immediately refreshes its metadata
- Search failures and empty results are shown separately, and newer searches replace outdated in-flight searches

### Impact
`✅ Manual recovery from incorrect or missing AniList matches`
`✅ Clearer distinction between similar manga entries`
`✅ Faster metadata correction after choosing an entry`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>


HeartlessVeteran2:
Five findings across @sourcery-ai and @codeant-ai. Four fixed in 4471eb5, one declined.

### The best of them: a failure with no message read as "no results"

`AniListPickerState.error` held `exceptionOrNull()?.message`, and `Throwable.message` is nullable. A failure carrying no message set `error` to null — which `isEmpty` then reported as an empty result. The picker would tell the user AniList has nothing under that title when in fact the search never completed.

That's bad on its own, and worse in context: the code immediately above it carries a comment saying a failed search and an empty one *"are different answers with different next steps, so they never share a message."* I wrote the principle and then broke it one line down by discarding the only thing distinguishing them.

`error` is now the `Throwable`. Non-null means failed regardless of what it carries, and the UI supplies a fallback string. There's a test for the null-message case specifically.

### `displayTitle` could be blank — fixed, but not the suggested way

Sourcery's diagnosis was right: all three title fields null or blank yields `""`, and a picker row with no visible title is a tappable target you can't identify.

I didn't take the suggested `?: "(Untitled)"`, because that puts user-facing prose in a domain model — this repo keeps strings in `strings.xml`, and the domain module has no resources. Instead `displayTitle` is now `String?`, which makes the absence explicit and forces the caller to supply a placeholder from a string resource. Sourcery's other half — try synonyms before giving up — I did take: they're real names, and they're already placeholder-free by the time a candidate exists.

### Cover placeholder, dismiss-cancels-search, blank-query-clears

All three valid, all three fixed. On the dismiss one I'll be precise about what it is: **housekeeping, not a correctness fix.** The search's state update already no-ops once the picker is gone (`state.anilistPicker ?: return@update state`), so nothing stale could land. What it buys is not holding a request against a rate limit that can make the next one wait 90 seconds.

The blank-query one was a real inconsistency — deleting the text and pressing Search left the previous query's candidates under an empty box, and the list is the part that gets tapped.

### Declined: cancelling the search when the query *text* changes

@codeant-ai filed this as a race condition. I don't think it is one, and the distinction matters for what the picker promises.

Search here is **submit-driven**, not type-to-search. Results reflect the last *submitted* query; the text field is a draft the user is composing. So a search completing while the box has been edited isn't stale data racing fresh data — it's the answer to the question that was actually asked. Cancelling on keystroke would mean editing the text silently discards results you asked for and might still want.

On the stated risk — "users may select an unintended AniList candidate" — a candidate row shows its title, cover and year, and tapping one links the entry you can see. There's no path where a query/results mismatch causes an action that differs from what was on screen.

If the intent is type-to-search semantics, that's a different design (needs debouncing so it doesn't issue a request per keystroke against this rate limit), and worth doing deliberately rather than by adding a cancel.

---

Full local set green: `:app:assembleDebug`, `testDebugUnitTest`, `:domain:test`, `detekt`, `ktlintCheck`. Three new tests, including the null-message failure and the blank-query clear.

---

HeartlessVeteran2:
@codeant-ai — took the closing note, in f983294. It was filed as a minor consideration and I think it's the better call outright, so `error: Throwable?` is now `searchFailed: Boolean`.

The reason to hold the throwable was that `Throwable.message` is nullable, so deriving the field from the message turned a failure carrying none into "no results" — the exact distinction the picker exists to preserve. But holding it *invited rendering it*, which is what you spotted: "Unable to resolve host graphql.anilist.co" is written for a developer, not someone trying to link a manga.

A flag keeps the distinction and removes the temptation in one move. There's nothing left to render, so nothing developer-facing can reach the screen later because someone decides the detail would be helpful. Same reasoning as `cachedMetadata`/`metadata` earlier in this stack: make the safe thing the only thing, rather than the thing you have to remember.

Nothing is lost by it. The metadata endpoint is unauthenticated, so no message could have carried a credential — this was a UX and habit problem rather than a leak — and with no logging sink in this app the exception had no second consumer to justify keeping it.

Tests updated; the message-less-failure case still pins that a failure never reads as an empty result.

---

HeartlessVeteran2:
Both of the latest findings were about tests asserting less than they claim. Fixed in a1327de.

### @codeant-ai — cancellation vs `Result.failure`

Checked, and the answer is no: `catch (CancellationException) { throw e }` precedes the general catch in `AniListSearchRepositoryImpl`. But **nothing pinned it**, and the ordering is the only thing holding it — `kotlinx.coroutines.CancellationException` is a plain `Exception`, so swapping the two blocks would swallow it silently.

The consequence is worse than the surface reading. The picker cancels an in-flight search whenever the query is replaced or the dialog is dismissed. If cancellation came back as a failure, the caller would continue to its next statement — and **`_state.update` is not a suspension point**, so a cancelled coroutine still executes it. Every dismiss would flash "Couldn't search AniList" at the user.

So this was an invariant with no protection, one catch-block reorder away from a visible bug. Test added, confirmed to fail with the rethrow deleted. Good catch — it was framed as a minor consideration and it was the sharpest thing on the PR.

### @cubic-dev-ai — "picking closes the picker" asserted against a picker that was never open

Correct and embarrassing. `picker_selectingACandidateStoresAUserLinkAndForcesARefetch` dispatched `SelectAniListCandidate` without ever opening the picker, so `anilistPicker` was already null and the assertion was true *before* the event — it would have passed with the close deleted.

Now it opens the picker, asserts the precondition, then selects. Confirmed to fail when the close is removed.

---

Worth naming the pattern, since it's four PRs running. Every real finding in this stack has been one of two shapes:

1. **A comment or name asserting a property the code has on one path and not another** — the empty-list contract in #1240, the `waitsForCachedSynonyms` test name in #1241, `error` losing the failed/empty distinction here.
2. **An assertion that would pass with the thing it names deleted** — the guess test in #1241, the concurrency test whose fake DAO delayed on the wrong side of the read, and this one.

Neither is a subtle-logic failure. Both are cases where I wrote the correct rule and then didn't check that the code or the test actually held to it. The removal check (delete the guard, watch a test fail) catches shape 2 reliably and I've been running it — it caught the fake-DAO problem — but I only run it on guards I *notice* are guards. The close-on-pick was one I didn't classify as a guard at all.

---

HeartlessVeteran2:
@codeant-ai — correct, and I took the second branch you offered: the docs and tests change, not the behaviour. Fixed in 37b750a.

You quoted the exact mismatch. The KDoc said the previous search is cancelled because *"the user retyping means the previous query is no longer what they want"*, while the code cancels on **submit**. And submit-driven is the behaviour I argued for two comments earlier when declining cancel-on-keystroke — so the code was right and the prose was describing the design I had just rejected.

That's the fifth instance in this stack of a comment asserting something adjacent to what the code does, and the first where I'd already written the correct version elsewhere **in the same conversation**. Which sharpens the diagnosis: it isn't that I don't know the rule, it's that the prose gets written from the intent in my head rather than re-derived from the code in front of me, and nothing forces the comparison.

Three changes:

- The KDoc now states the contract directly — results answer the last query actually *submitted*, the text field is a draft, only submitting again supersedes a running search — and names type-to-search as the different design it would be, with the debouncing it would need.
- `picker_aRetypedQueryReplacesTheInFlightSearch` had the same mismatch in its name. It dispatches an edit *and* a submit, so it was never testing what "retyped" implies. Now `picker_resubmittingReplacesTheInFlightSearch`.
- **The half nobody had pinned** — that editing alone does *not* cancel — now has its own test. You were right that the existing one couldn't catch it, because dispatching both events back to back never exercises the interval between them. Confirmed to fail when `SetAniListPickerQuery` is made to cancel the job, which is the check that makes it worth having.

That last one is the part I'd have missed on my own. The contract had two halves, one was tested, and the untested half was the one the documentation got wrong — those are related, not coincidental: prose is where a claim goes when no test is holding it.

---

PR #1244, opened 2026-08-08
Make reading a chapter actually advance tracker progress

HeartlessVeteran2:
## **User description**
Stage 5d — and it turned out to be a bigger fix than the plan assumed.

The plan said the reader→tracker wiring *"already exists; this tightens the trigger condition."* I went to tighten it and found the wiring is disconnected at the join.

## Progress recorded by reading never reached the push

`recordLocalChange` wrote `localLastChapterRead` on the **sync-state row**. `syncManga`'s local-wins branch builds what it sends from **`trackRepository.getEntry(...)`** — the `TrackEntry` — and nothing on the read path ever touched that. Grepping `localLastChapterRead` in `TrackerSyncRepositoryImpl`, it's written by `recordLocalChange` and read in exactly one other place: the conflict check.

So finishing a chapter marked the sync PENDING, pushed the *previous* entry unchanged, and then marked itself SYNCED. **Reading never advanced progress on any tracker**, and the failure was invisible because the sync reported success.

It also explains something I'd have found later in Stage 8: the per-tracker chips added in #1239 never move as you read, because they render the same `TrackEntry`.

`recordLocalChange` now writes the entry too — the thing the push actually reads.

## And progress only ever goes forward

Re-reading an old chapter is normal, and must not report to AniList that you've un-read fifty chapters. Both writes take the maximum.

The sync-state row needs the same guard for a **second, separate reason**: `localLastChapterRead` feeds conflict detection, so a re-read writing a lower number could disagree with an unchanged remote and hand the user a resolution prompt for a conflict they created by opening chapter 5. That one surprised me — the guard is load-bearing in two different subsystems.

Status is recorded either way, since dropping a manga is a real change even when no chapter was read. That's why the timestamp and PENDING marking sit outside the progress guard, and there's a test pinning it.

## Scope

The **95%-of-chapter completion trigger** from the plan is deliberately not here. It's a behaviour change with no correctness argument behind it — stopping four pages from the end of a hundred-page chapter would start counting as read — and I'd rather it not ride along with a data-integrity fix. Happy to do it next if you want it; it's small.

## Verification

Five tests, none of which existed before. Two of them fail against the original code:

- `reading a new chapter advances the entry the push actually sends`
- `re-reading cannot manufacture a conflict on the sync state either`

The other three pin the no-regression rule, the status-still-records case, and the no-local-entry case.

---

## Summary by Sourcery

Ensure chapter reading updates tracker progress consistently while preventing regressions or spurious conflicts.

Bug Fixes:
- Connect chapter-reading progress updates to the TrackEntry used for sync so tracker progress and UI chips advance when reading.
- Guard progress updates in both the TrackEntry and sync-state row so re-reading cannot move tracker progress backwards or manufacture conflicts.

Tests:
- Add unit tests covering chapter-to-tracker hand-off, forward-only progress updates, conflict prevention, status-only changes, and behavior when no local entry exists.

---
## Summary by cubic
Make chapter reads advance tracker progress and update chips. Sync is now race-safe across auto and manual paths: mid-flight reads aren’t lost, unlinks don’t resurrect, progress never goes backward, and rows don’t get stuck SYNCING.

- **Bug Fixes**
  - Upsert the `TrackEntry` in `recordLocalChange`; use max(current, incoming) for progress; always record status; works even before a sync-state row exists.
  - Serialize all whole-row sync-state writes with a per-row striped lock; re-check after network calls using row id to detect relinks or deletes; keep rows PENDING and do not advance `lastSuccessfulSync` when a local change lands; in remote-wins, leave the pre-fetch remote snapshot so the next sync sees a conflict; applies to auto sync, “Sync Now”, push/pull, and conflict resolution.
  - Handle unlink cleanly: delete the matching sync-state row on unlink and treat “unlinked during sync” as a failure to prevent orphan retries, stale remote IDs/high-water marks, and link resurrection.
  - Never leave rows stuck in SYNCING: use targeted `markSyncError` updates to set ERROR on returned failures; pushing with no local entry now returns a failure instead of silent success.
  - Add tests for concurrency (mid-flight reads, double writes), no-op sync safety, unlink cleanup and re-link, relink-mid-sync identity, forward-only updates, status-only updates, first-sync (tracked-but-never-synced), and missing-entry push.

<sup>Written for commit 56c36c712e21952f6942be566d8d9430ea5af520. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/HeartlessVeteran2/Otaku-Reader/pull/1244?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

___

## **CodeAnt-AI Description**
Advance tracker progress when chapters are read

### What Changed
- Reading a chapter now updates the tracker progress that is sent during synchronization
- Re-reading an earlier chapter no longer moves progress backward or creates a false sync conflict
- Status changes still sync even when chapter progress does not increase
- Added coverage for progress updates, rereads, status changes, and missing tracker entries

### Impact
`✅ Reading advances tracker progress`
`✅ Fewer false sync conflicts from rereading`
`✅ Tracker status changes remain synchronized`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>


HeartlessVeteran2:
Both cubic findings were real. Fixed in 75a5de3.

**P1 — a read during an in-flight sync was lost, not just delayed.**

`syncManga` read the sync-state row, made the network call, then wrote `syncState.copy(...)` — a copy of the row as it looked *before* the request. A `recordLocalChange` landing in that window got overwritten.

The part that made it silent rather than merely late: `localChanged` is `localLastModified > lastSuccessfulSync`. Marking the row SYNCED advances `lastSuccessfulSync` past the new modification stamp, so the *next* sync also decides there is nothing to send. The chapter never gets pushed and nothing anywhere reports a failure.

`localChangeSince(snapshot)` re-reads the row after the network call and returns it only if `localLastModified` moved. Both write branches build their update from that fresher row, and when it moved they leave the status `PENDING` with `lastSuccessfulSync` untouched, so the next pass picks the change up. Compared on the timestamp rather than the chapter number because a status-only change matters here too and would not move the number.

The remote-wins branch needed more than a status flip: once a local change has landed, that is no longer a clean pull, it is a conflict. It now records the remote values, leaves the row `PENDING`, and does *not* overwrite the local entry — adopting the remote value there would discard the chapter the user just finished in favour of a value that predates it.

**P2 — the push reported success after pushing nothing.** With no local entry it returned `SyncResult(true, "Sync successful")`. It now returns the same `"No local entry found for manga"` failure the auto-create path already returns for that condition.

**Complexity.** Inlining the guard put `syncManga` over detekt's `CognitiveComplexMethod` threshold. Extracted `pushLocalToTracker` / `applyRemoteToLocal` rather than suppressing — the rule was right that the method had grown past readable.

**Verification.** `TrackerSyncConcurrentReadTest` covers the interleaving; three of its five cases fail when `localChangeSince` is stubbed to return `null`, so they are testing the guard and not the happy path around it.

---

HeartlessVeteran2:
Both findings from the latest review were right, and tracing the first one turned up a third bug. All in `047a112`.

**P1 — the re-read narrowed the window, it didn't close it.**

Correct, and it's the case this repo's own review checklist warns about: *a liveness check and the write it guards belong inside the same lock, not each taking one separately.* `localChangeSince` and the `updateSyncState` that follows are two separate suspending calls, so a `recordLocalChange` landing between them is still lost.

A striped mutex keyed on `(mangaId, trackerId)` now spans each read-decide-write tail. It is **never held across a network call** — `tracker.update()` runs outside it — so a slow tracker can't block a chapter from being recorded. Striped rather than a map of per-row mutexes, matching `AniListMetadataRepository`: bounded by construction, and a stripe collision costs a little parallelism and can never cause a wrong result. It serializes writers *within the process*, which is every production writer of this table; the KDoc says that rather than implying a database-level guarantee it doesn't provide.

**P2 — the bail-out was suppressing the exact signal it existed to preserve.**

Also correct, and the trace is precisely as described. `remoteChanged` compares the fetched value against the stored remote snapshot, so recording the fetched value made the next sync compare it against itself, find them equal, and conclude the remote hadn't moved. Only `localChanged` stayed true → clean local push → the remote chapter overwritten, no conflict prompt, no error.

The branch now leaves the pre-fetch snapshot alone and writes nothing at all; `recordLocalChange` has already left the row PENDING with a fresh timestamp, so there was nothing to write in the first place.

This is defect shape #1 from `CLAUDE.md` — the comment asserted a property ("the next sync re-evaluates it as the conflict it now is") that the code had on no path. Worth noting the comment is what made it invisible: it read convincingly enough that I didn't re-derive it.

**The third bug, found while checking the earlier "no local entry" finding.**

That state isn't hypothetical, and the route to it is a leak. `unlinkManga` deletes the `TrackEntry`; nothing deletes the `tracker_sync_state` row. There's no foreign key between the two tables, and `deleteSyncStateForManga` had **no production caller at all**. The orphan is not inert:

- `recordLocalChange` still finds it and keeps marking it PENDING, so `syncAllPending` keeps retrying a tracker the user deliberately unlinked — and with the P2 fix from the previous round, every one of those passes now *fails*, with an error that can never clear.
- `syncManga` auto-creates a row only when it finds none, so re-linking the same manga to a **different** series reuses the stale row — carrying the old `remoteId` and chapter history, and pinned to the old high-water mark by the `maxOf` in `recordLocalChange`.

`deleteEntry` now removes both, at the single choke point for entry removal so a future caller can't forget. Scoped to the one tracker — `deleteSyncStateForManga` would silently unlink MAL and Kitsu too, and there's a test pinning that.

(This rode along in `047a112` rather than getting its own commit — the message covers the other two. Flagging it so it isn't missed reading the log.)

**On the tests.** Five new ones, each confirmed to fail when its own fix is reverted. Both new ones needed their fakes rebuilt before they were worth anything:

- The conflict test first stubbed `getSyncState` with `returnsMany`, so pass 1's write went into a mock void and pass 2 read a canned row — it passed with the bug fully reinstated. It's stateful now.
- The lock test yielded *before* reading rather than snapshotting first, so the late read observed the other coroutine's write and the interleaving never appeared. Same shape of mistake as the DAO fake earlier in this branch; the comment now says why the order matters.

Gradle's build cache initially hid both — a mutation I'd already built once came back `FROM-CACHE` as a pass. `--rerun-tasks` for mutation checks from here on.

---

HeartlessVeteran2:
All three real. `5c1badf`.

**P1, the unlink resurrection.** Traced and confirmed. `applyRemoteToLocal` upserts `localEntry ?: remoteEntry` — a fallback meant for "tracked on the service, no local row yet". After `deleteEntry`, `getEntry` also returns null, so the fallback fires and recreates the entry from the fetched remote value. The manga reappears as tracked moments after the user removed it. (The `updateSyncState` beside it is a no-op on a deleted row, so the sync row stays gone and only the entry comes back — which is why it would present as a half-restored link.)

The root cause is that the re-read was a nullable "did it change", which cannot express a third state. It's now `RowSince`: `Unchanged` / `Changed` / `Gone`, and `Gone` aborts before the upsert. Worth noting your suggested alternative — coordinating the delete with the sync lock — would also work, but the abort is better here: it needs no cross-class lock sharing, and the sync-state row's absence *is* the unlink signal, so there's nothing to keep in step.

**P1, the no-op branch.** Confirmed. `markSyncSuccess` sets SYNCED and advances `lastSuccessfulSync`; `localChanged` is `localLastModified > lastSuccessfulSync`, so applying it to a row a read had just marked PENDING means the next sync sees nothing to send. Same lost update as the push and pull paths, reached by a route I hadn't covered. `completeNoOpSync` takes the lock and re-reads first.

**P2, the KDoc.** You're right, and it was wrong in the way that matters. I checked `pushToTracker`, `pullFromTracker` and `resolveConflict`: each carried its own copy of these writes, from a pre-request snapshot and unlocked — so pressing Sync Now, or resolving a conflict, still lost a concurrent read.

Rather than narrow the comment, I removed the reason it was false. All three were near-verbatim duplicates of `pushLocalToTracker` / `applyRemoteToLocal` (five copies of two operations), so they now delegate to them. `resolveConflict`'s error path was writing the whole snapshot back too; it uses a targeted `markSyncError` now.

The KDoc states the invariant that actually holds and is checkable: every **whole-row** write goes through the lock, and the remaining writers are targeted `@Query` column updates that cannot touch `localLastChapterRead` or `localLastModified` — plus a note that a column added to one of those statements needs re-checking against that property. I verified it by enumerating every `trackerSyncDao` write in the file against its enclosing lock rather than by re-reading the prose.

That's twice now on this PR that a comment of mine asserted a property the code had on only some paths. The pattern I'm taking from it: a comment claiming coverage over a *set* of call sites is worth nothing unless I've enumerated the set — the sweeping phrasing ("every production writer") is itself the tell.

Two tests, each confirmed to fail with its own guard reverted, using `--rerun-tasks`.

---

HeartlessVeteran2:
Both real. `56c36c7`.

**Row identity.** Confirmed, and the reason it's easy to miss is that `getSyncState`'s key is *reused*: `(mangaId, trackerId)` is a unique index, not an identity. Unlink + relink during one request deletes the row and inserts a fresh one under the same key with a new autogenerated id, so the lookup succeeds and the new row's `localLastModified` is naturally newer than the snapshot's. `Changed` then hands the push branch a row belonging to a different link, which stamps it with the old link's tracker response *and* the old row's `lastSuccessfulSync`. Took your suggestion as written — a differing id is `Gone`, since the snapshot's row genuinely is.

**Stuck SYNCING.** Confirmed, including that it's terminal: `getPendingSyncs` selects `syncStatus = PENDING`, so a row left at SYNCING is never picked up by `syncAllPending` again. The manga just stops syncing and the only symptom is a spinner.

I didn't take the suggested diff, and the reason is worth stating. Restoring the no-entry check in `pushToTracker` fixes the instance you pointed at, but I enumerated the failure returns that land after a SYNCING stamp and there are six, across three methods — `syncManga` has the identical hole on "Invalid remote ID" and "Entry not found on tracker", and `pullFromTracker` on the latter. Two of those predate this PR. Patching the one in front of me is exactly what left three sibling paths unlocked in the previous round, so this is a `failSync()` helper that records the message through the targeted `markSyncError` and returns the result.

ERROR rather than PENDING deliberately: none of these failures fix themselves, so requeuing would retry the same failure on every pass.

Two tests, each confirmed to fail with its own guard reverted.

Running count on this PR: five review rounds, eleven findings, all eleven real. Every one has been a variant of the same lost-update, reached by a path the previous round's fix didn't cover — which is a fair signal that the sync state machine wants a single guarded write path rather than seven entry points that each stamp and clear status by hand. That's a larger refactor than this PR should carry, so I'm not doing it here, but it's the honest read on why this keeps recurring.

---

PR #1249, opened 2026-08-08
Add characters and staff to the AniList details section

HeartlessVeteran2:
## **User description**
Slice 4d, first half — the cast and credits carousels on the details screen.

Relations and external links are a different data shape and different UI, so they follow in their own PR rather than riding along here.

## The one real decision: how to store them

Everything currently in `MangaMetadataEntity` is a scalar or a `List<String>`, and the string lists are stored delimited. Tags are the awkward case already — `tagNames` and `tagRanks` are two columns that **must stay the same length**, and the mapper has to zip-and-drop to survive them disagreeing. The entity KDoc flags it.

That's tolerable for two fields. A person has four (`id`, `name`, `imageUrl`, `role`) and there are two such lists, so the same encoding means **eight columns and eight chances for that invariant to slip**. So: one JSON column per list, via a `kotlinx.serialization` TypeConverter. Each record stays whole — a field can't go missing independently of the person it belongs to.

Not a child table, for the same reason tags aren't one: the access pattern is "read every person for this manga, always together, never queried across manga", so a join buys nothing and costs a second table plus its own cascade.

**The converter returns an empty list on a parse failure rather than throwing.** That's a judgement call and it's only safe here: `manga_metadata` is a disposable seven-day cache with an upstream that can always be re-fetched. A Room read that throws wouldn't degrade one section — every screen observing this manga collects a Flow off that read. The KDoc says explicitly not to copy it to a table that owns its data, where swallowing a parse error turns corruption into silent data loss.

## `role` is stored raw

For a character it's an AniList enum — `MAIN`, `SUPPORTING`, `BACKGROUND`. For staff it's free text AniList keeps verbatim, like "Story & Art". Both go into the cache unchanged and get formatted at render, by the caller, because only the caller knows which list it's holding.

Prettifying the enum on the way in would turn `MAIN` into "Main" — indistinguishable from a staff credit that genuinely reads "Main", at which point nothing downstream can tell the two conventions apart. Same reason `PersonCarousel` takes a `roleLabel` lambda instead of sniffing the values.

## Two bugs the tests caught before they shipped

Both would have compiled cleanly and failed later, away from the cause:

**`@Serializable` with no serializer.** I added `kotlinx-serialization-json` to `core:database` but not the compiler plugin. The annotation resolved, the module compiled, and `StoredPerson` had no generated serializer — it would have thrown `SerializerNotFound` on the first read of a cached row. Caught by the converter round-trip test.

**The `$` that goes the wrong way.** `METADATA_QUERY` is a raw string containing two kinds of dollar. `${'$'}id` must survive into the sent query as a *literal* dollar, because it's a GraphQL variable the server binds. `$PEOPLE_PER_PAGE` is a Kotlin template that must be substituted before sending. I wrote the second in the first's escaped form — which reads correctly at a glance and ships `perPage: ${PEOPLE_PER_PAGE}` to AniList. `AniListMetadataQueryTest` now pins both directions.

## Schema v43

`ALTER TABLE manga_metadata ADD COLUMN characters/staff TEXT NOT NULL DEFAULT '[]'`.

Both columns declare `@ColumnInfo(defaultValue = "[]")` on the entity, not just the default in the migration. Room compares the default it expects against the one SQLite reports, so a default living in only one of the two places is a validation failure — **on upgrade only, never on a fresh install**, which is precisely how that class of bug reaches users who already have the app. `runMigrationsAndValidate` is the assertion that catches it, and there's a test confirming a pre-existing cached row backfills to something the converter can actually parse (`''` would work by accident; `'{}'` would not).

## Verification

15 new tests. Every new guard was deleted and confirmed to fail its own test, with `--rerun-tasks` — Gradle's build cache served a stale pass for an already-built mutation earlier in this branch, so cached mutation runs can't be trusted.

| Mutation | Test that failed |
|---|---|
| Keep nameless people | `an edge with no usable name is dropped, but a missing image is not` |
| Escaped Kotlin template | `the page size is substituted, not sent as a literal template` |
| Converter rethrows | `a malformed blob yields an empty list instead of propagating out of a Room read` |
| Migration default `'{}'` | `migration42To43_validatesAgainstTheExportedSchema` |

## Not here

Relations, external links, and the "people you follow are reading this" section from the plan. Each is a separate shape and a separate UI; splitting keeps one reviewable idea per PR.

---

## Summary by Sourcery

Add AniList characters and staff metadata to the manga details screen, storing people as JSON in the cache and migrating the database schema to v43.

New Features:
- Display character and staff carousels in the AniList details section using a shared person model and a reusable `PersonCarousel` composable.
- Request characters and staff, including roles and images, from AniList in the metadata query and expose them through the domain `MangaMetadata` model.

Enhancements:
- Store cast and staff lists as JSON blobs in `manga_metadata` via a `StoredPerson` type and Room type converters instead of parallel delimited columns.
- Document and pin the behaviour of the person list converter, including handling of malformed or forward-compatible JSON and unknown fields.

Build:
- Enable the Kotlin serialization plugin and add `kotlinx-serialization-json` dependency in the core database module.

Documentation:
- Update CLAUDE.md to describe the extended AniList metadata section with characters and staff carousels and the new schema version v43.

Tests:
- Add unit tests for the AniList metadata repository to verify character/staff mapping, ordering, and behaviour when entries or lists are missing.
- Add migration tests for v42→v43 to validate new `characters` and `staff` columns, their defaults, and schema consistency on upgrade.
- Add tests for the person list JSON converter and for the AniList metadata query string to ensure correct interpolation and requested fields.

Chores:
- Export the Room schema for version 43 and update database version constants and migration chain tests.

---
## Summary by cubic
Adds characters and staff carousels to the AniList details section. Fetches, caches, and displays cast and credits with clear role labels, while hardening decoding and UI to avoid crashes.

- **New Features**
  - Extend AniList `METADATA_QUERY` to fetch characters and staff (25 per page) and render each as a carousel.
  - Cache people as JSON in `manga_metadata` (`characters`, `staff`) via a `kotlinx.serialization` TypeConverter; schema v43 adds both columns with `DEFAULT '[]'`.
  - Store `role` as-is; UI prettifies character enum roles and shows staff roles verbatim.

- **Bug Fixes**
  - Prevent duplicate-key crashes in the carousels by removing the `id`-based `LazyRow` key; preserve multiple staff credits and dedupe only exact duplicates in the mapper.
  - Narrow JSON decode handling to `SerializationException` and surface discards via `DatabaseConverters`’ `onDecodeFailure` hook (logs payload length without content).
  - Add Kotlin serialization compiler plugin to `core:database` and `kotlinx.serialization.json` runtime to generate serializers.
  - Fix `$` templating in `METADATA_QUERY`: keep `${'$'}id` as a GraphQL variable and substitute `$PEOPLE_PER_PAGE` in Kotlin.

<sup>Written for commit 552e991e155204cbff5d14e1ab15970348b613e1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/HeartlessVeteran2/Otaku-Reader/pull/1249?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

___

## **CodeAnt-AI Description**
Add character and staff carousels to AniList manga details

### What Changed
- AniList details now show up to 25 characters and staff members in separate horizontal carousels with portraits, names, and roles.
- Character roles are displayed as readable labels, while staff credits retain AniList’s original wording.
- Entries without a usable name are omitted; people without portraits remain visible.
- Cast and credit data is cached across refreshes and older database rows are upgraded with empty lists.
- Invalid cached people data no longer prevents the manga details screen from loading and can be repaired on the next refresh.

### Impact
`✅ Cast and staff visible on manga details`
`✅ Clearer character roles and staff credits`
`✅ Details screen survives malformed cast or credit cache data`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>


HeartlessVeteran2:
All five points addressed in `552e991e` — four accepted, one declined with reasoning. Thanks both; the LazyRow key one in particular was a live crash.

**The duplicate key (cubic P2) — real, and worse than a theoretical collision.** AniList returns **one staff edge per credit**, so a person responsible for both Story and Art is two legitimate edges sharing one person id. Not a hypothetical malformed-data case: it's the normal shape for the most common manga credit there is. `LazyList` throws `IllegalArgumentException` on a repeated key, so this would have taken down the entire details screen.

I didn't take the suggested `distinctBy { it.id }`, because it fixes the crash by **discarding the second credit** — the reader loses "Art" for someone also credited "Story". Instead:

- The list is **unkeyed**. My justification for the key ("recomposition survives a wholesale replacement") doesn't hold up: these cards are read-only, hold no internal state, never reorder and never animate, so positional keying is both correct and free. The key was solving nothing and could crash.
- The mapper still dedupes, with `distinct()` rather than `distinctBy` — it collapses only edges identical in *every* field, which carry nothing the first copy doesn't, while two genuine credits survive.

Both directions are pinned, and I checked both bite: swapping in `distinctBy { it.id }` fails `a person credited for two roles keeps both credits`; removing dedup entirely fails `an entirely duplicate edge is collapsed`.

**`runCatching` too wide (cubic P2) — accepted.** Narrowed to `SerializationException`. Masking an `OutOfMemoryError` or `StackOverflowError` as "this manga has no characters" while the process is in trouble is exactly the kind of thing the swallow was never meant to cover.

One correction on the reasoning, since it matters for the general rule: the `CancellationException` mechanism described can't occur here. `decodeFromString` on a `String` is pure synchronous work with no suspension point, and cancellation isn't injected into arbitrary synchronous frames — it's delivered at suspension points. The `Error` cases are the real ones, and they're sufficient on their own.

**Silent discard (Sourcery) — accepted, and it's the sharper version of the same worry.** One row failing is a curiosity; *every* row failing means an incompatible change to `StoredPerson`, and nothing else would report it, because Room validates a table's columns and not the shape of JSON inside a TEXT one. The carousels would vanish app-wide with no other symptom.

`DatabaseConverters` now takes an `onDecodeFailure` with a default — Kotlin generates the parameterless constructor Room needs when every parameter has one, so Room's construction is untouched while a test can inject a recorder and assert the signal actually fires. That test fails if the `onFailure` is deleted, which a "returns empty" assertion would not.

Two deviations from the suggested implementation, both deliberate:

- **Warning, not debug.** This project's `Logger.d` compiles out in release builds, so a debug-level diagnostic would be absent in precisely the situation it exists for. It's rare enough that it can't become spam — and if it isn't rare, that's the thing worth knowing.
- **Length, not payload.** The blob holds character and staff names for one specific manga, which identifies the title. What someone reads doesn't belong in logcat. Length plus the exception already separates truncation from a shape change, which is what a diagnosis turns on.

I also skipped the suggested `value == "[]"` fast path — `decodeFromString("[]")` already returns an empty list, so it's a branch that restates existing behaviour.

**`PERSON_ROW_EDGE_PADDING = 0.dp` (cubic P3) — accepted, removed.** Leftover from an inset I decided against.

**The literal `25` in the query test (Sourcery) — declined.** A test that reads its expectation from the same constant as the code under test can't distinguish a correct page size from a wrong one: bump `PEOPLE_PER_PAGE` to 500, blow AniList's complexity budget, and the test still passes having verified only that Kotlin can interpolate. Restating the number makes changing it a deliberate act that touches the file where the reason for the limit is written down. (It's also the only option — the constant is file-private.) I've added a comment so it doesn't read as an oversight to the next person.

Full verification green. Every new guard deleted and confirmed to fail its own test, with `--rerun-tasks`.

---

PR #1250, opened 2026-08-08
Add related manga and external links to the AniList details section

HeartlessVeteran2:
## **User description**
Slice 4d, second half — the remaining two AniList metadata lists. Storage mechanism is the one established in #1249 (JSON column per list via a TypeConverter), so this PR is mostly about two decisions that needed making rather than new infrastructure. Schema v44.

## Relations are filtered to manga, at the boundary

AniList lists a manga's **anime adaptation** among its relations. This app reads manga and novels and has no anime surface at all — so an adaptation tile could only sit there doing nothing when tapped.

That matters because *every* relation tile is tappable: the point of showing a sequel is being able to go and find it. Filtering at the mapper means the UI never has to ask whether a given tile is actionable, which is the version of this that can't drift — the alternative (render everything, make only some tappable) is a rule the next carousel would have to remember.

## Tapping a relation searches; it doesn't navigate

A relation is an **AniList media id with no local record**. It may not be in the library, and it may not exist on any source the user has installed. There's nothing to navigate to, and inventing a details screen for a manga the app doesn't have is a much larger concept than this slice.

So a tap reuses `NavigateToGlobalSearch` — the same effect the tag chips in this section already use, and the same one `DetailsViewModel` already uses for a title suggestion. It lands the user somewhere they can actually act.

## The URL scheme is checked twice, and that's deliberate

AniList's `externalLinks` are **user-submitted third-party data**. Two checks, doing different jobs:

- **In the repository**, deciding what's worth caching. A `javascript:` URL stored now is a chip that can only be refused later.
- **In `ExternalLinkChips`**, before a URL becomes an `Intent`.

The second isn't redundant. The cache is a table on the device: a row written by an older build, or restored from a backup, **never passed through today's mapper**. The check at the point where a URL turns into an Intent is the one that can't be bypassed by where the data came from. Both are documented as mirroring each other so neither reads as an oversight.

A chip failing the check isn't rendered at all, rather than rendered and refused on tap — a control whose only behaviour is to reject the user is worse than an absent one.

## Smaller things

- `relationType` is stored raw and prettified at render, matching the rule set for a person's `role` in #1249.
- A link with no `site` falls back to its host, because AniList leaves `site` blank far more often than `url` — a nameless chip is still usable, a discarded one isn't.
- The three JSON list converters now share one body. The interesting part is *which* exception is tolerated (`SerializationException`, not `runCatching` — settled in #1249), and three hand-written copies would be three chances to widen one back without noticing.
- The relations carousel is unkeyed, following the crash found in #1249. Relations don't repeat an id the way staff credits do, but a static read-only row gains nothing from keys and that's the failure mode that already took down this screen once.

## A trap worth recording

Compiling after adding the entity fields but **before** bumping the DB version rewrites the *previous* version's exported schema JSON. `43.json` came out describing v44's shape, and the resulting failure looks like a broken migration (`duplicate column name: relations`) rather than a corrupted fixture. Restored from the merged commit; the note is in the commit message so the next schema bump doesn't lose an hour to it.

## Verification

13 new tests. Every new guard deleted and confirmed to fail its own test, with `--rerun-tasks`:

| Mutation | Test that failed |
|---|---|
| Drop the `type == MANGA` filter | `an anime relation is dropped and a manga relation is kept` |
| Drop the scheme filter | `only http and https links are cached` |
| Drop the host fallback | `a link with no site name falls back to its host` |

Plus migration tests for v43→v44 (column existence, `'[]'` backfill of pre-existing rows, and `runMigrationsAndValidate` against the exported schema), and converter round-trips including one asserting all three list converters share the same discard policy.

## Not here

The "people you follow are reading this" section from the plan. It needs an authenticated viewer query rather than the public media query this file uses, so it belongs with the tracker work rather than the metadata cache.

---

## Summary by Sourcery

Add AniList-related manga and external links to the metadata cache and details screen, including schema bump and storage/mapping support.

New Features:
- Expose related manga relations from AniList in the domain model and render them as a tappable carousel on the details screen.
- Expose AniList external links in the domain model and render them as link chips that open the destination in a browser.

Enhancements:
- Filter AniList relations to manga-only and require titles at the repository boundary so only actionable relation tiles are stored and shown.
- Validate and normalise AniList external links when caching, including http/https scheme enforcement and host-based site name fallback.
- Factor JSON list decoding for people, relations, and external links into a shared converter body with a consistent failure/discard policy.

Build:
- Bump Room schema version to v44 and document the new relations/externalLinks columns and migration behaviour in the project guide.

Documentation:
- Update CLAUDE documentation to describe the expanded AniList metadata section (relations carousel and external link chips) and the new schema version.

Tests:
- Extend repository, converter, and migration tests to cover relations and external links mapping, storage round-trips, URL filtering, host fallback, and the v43→v44 migration chain.

---
## Summary by cubic
Adds related manga and external links to the AniList details section. Users can browse sequels/side stories and open safe off‑site links directly from the manga page.

- **New Features**
  - Related manga carousel (manga‑only). Tap runs a global search for the title.
  - External link chips (http/https only). Cache and launch share `isBrowsableHttpUrl()` and one hostname extractor; links must resolve to a real hostname and malformed authorities are hidden (rejects scheme‑only, authority‑only like `https://@/`, port‑only, and non‑numeric ports like `host:abc`). Chip labels use `browsableHostOrNull()` with IDN support validated via `IDN.toASCII` (STD3 rules), lowercase host, `www` trim, and userinfo/port stripped; the predicate and label derive from the same extractor to stay in sync.
  - Polish: `relationType` prettified; relation titles fall back userPreferred → romaji → english → native; carousel unkeyed.

- **Migration**
  - Schema v44: add `relations` and `externalLinks` JSON columns to `manga_metadata` with default `[]`.
  - JSON list converters for people/relations/links share one body with a strict discard policy.

<sup>Written for commit 933b44cc5bab0ec2c5415e9b17fb36ec90d9c34d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/HeartlessVeteran2/Otaku-Reader/pull/1250?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

___

## **CodeAnt-AI Description**
Add related manga and external links to AniList details

### What Changed
- AniList details now show related manga with covers, relationship types, and formats; anime adaptations and incomplete entries are excluded.
- Tapping a related manga searches for its title across the user's sources instead of opening an unavailable details page.
- AniList external links appear as chips and open in the browser; only valid HTTP and HTTPS links are shown, with missing site names replaced by the link's host.
- Existing metadata is upgraded to store these lists without losing cached data, while malformed cached lists fall back safely to empty sections.

### Impact
`✅ Find sequels and side stories from AniList details`
`✅ Open official and publisher links from manga pages`
`✅ Fewer unusable or unsafe metadata links`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>


HeartlessVeteran2:
All three accepted, in `09fa0e7a`. The CodeAnt one was a real bug.

**Relations keyed on `userPreferred` alone (CodeAnt) — correct, and worth spelling out why.** `userPreferred` is *derived*: it resolves the viewer's title-language setting and defaults to romaji when unauthenticated. So an entry whose romaji is missing can leave it null while `english` or `native` are perfectly good — and the relation vanished from the carousel with no trace at all. Not even a blank tile; just absent.

It's also an inconsistency I introduced without noticing: the main media already requests all four title variants and builds a fallback in `buildSynonyms()`, and the relation node asked for one. Fixed as suggested — all four requested, first meaningful one wins, `userPreferred` still first because when it *is* populated it's the field that encodes the user's preference.

I went slightly further and applied `PlaceholderTitles.isMeaningful`, the same filter the synonym set uses. AniList emits `"?"` and `"N/A"` for missing titles, and a tile labelled `?` is worse than an absent one here: it's tappable, and searching for `?` finds nothing.

**Duplicated `isBrowsableHttpUrl` (Sourcery) — accepted.** I'd conflated two things in the PR body. Two *call sites* is the design (the cache boundary decides what's worth storing; the Intent site is the check a backup-restored row can't bypass). Two *implementations* was just duplication, and "future divergence in scheme filtering" is exactly the right worry — widen one, leave the other behind. One helper in `core:common` now, both call sites unchanged.

**`hostOrNull` hand-rolled (Sourcery) — accepted, with a different parser than suggested.** Not `android.net.Uri`: these tests are plain JUnit, and `Uri.parse` would throw "not mocked" and force Robolectric on a pure-mapping test. `java.net.URI` gets the same correctness on the JVM.

Worth noting the case that made this more than tidiness: sliced by hand, `https://user:pw@example.test/path` yields `user:pw@example.test`, which puts a username on screen as a chip label. Ports and query strings were the ordinary rest. There's a test named for that specific case.

Moving the scheme check to `URI` also changed its behaviour in one way I want to flag rather than bury: it now **fails closed**. A URL that won't parse is refused, where the slicing version would have let some through. For a check whose job is keeping `javascript:`, `intent:`, `file:` and `content:` away from an `Intent`, "I can't tell what this is" has to mean no — losing a malformed link a browser might have salvaged is the cheaper mistake.

Seven tests for the shared helpers, plus two for the title fallback. Both new guards deleted and confirmed to fail their own tests with `--rerun-tasks`. Full verification green.

---

HeartlessVeteran2:
All three real, fixed in `e2147654`. I verified each against `java.net.URI` directly rather than reasoning about it, and that turned out to matter — **the suggested fix for the first one would have introduced a worse bug.**

**P2, scheme without authority — correct.** `http:foo` and `https:///path` both parse and report an http scheme with `authority = null`. They were cached and rendered as chips that open nothing. My KDoc already claimed fail-closed; the code only checked the scheme. Textbook case of prose asserting a property the code has on one axis only.

**But the check has to be on `authority`, not `host`.** The suggested `uri.host != null` is the obvious reading and it's wrong here:

```
https://例え.テスト/p   →  scheme=https  host=null  authority=例え.テスト
```

`URI.getHost()` returns null for an internationalised name, so requiring a host would have **silently dropped every IDN link** — turning a cosmetic labelling problem into a missing link, which is strictly worse than the dead chip being fixed. There's a test named for that case, and I confirmed it fails under the `host != null` version specifically.

Same probe run for the record:

| input | scheme | host | authority |
|---|---|---|---|
| `http:foo` | http | null | null |
| `https:///path` | https | null | null |
| `https://例え.テスト/p` | https | **null** | 例え.テスト |
| `https://xn--r8jz45g.xn--zckzah/p` | https | xn--r8jz45g.xn--zckzah | same |
| `https://WWW.Example.TEST/p` | https | **WWW.Example.TEST** | same |
| `https://user:pw@example.test/p` | https | example.test | user:pw@example.test |

**P3, uppercase `WWW.` — correct.** That table's fifth row confirms `URI` does not normalise host case, so `removePrefix("www.")` left it. Fixed by lowercasing first, which subsumes the case-insensitive strip rather than needing a separate `startsWith(ignoreCase)` dance. Hostnames are case-insensitive so the normalisation is right on its own merits, and Kotlin's no-arg `lowercase()` is the locale-invariant mapping, so it's safe under a Turkish locale.

**P3, IDN labelled with the full URL — correct**, and it's the same root cause as the first finding. When `getHost()` gives up, the label now falls back to the raw authority with userinfo and port stripped by hand — doing what `getHost()` would have done. The port is only stripped when what follows the final colon is all digits, so an IPv6 literal keeps its address rather than being truncated.

Ten tests on the helper now. Each of the three guards deleted and confirmed to fail its own test with `--rerun-tasks`, including the `host`-instead-of-`authority` variant.

---

HeartlessVeteran2:
Real, fixed in `2e2f1c38`. Confirmed by probe:

```
https://@/              scheme=https  host=null  authority=[@]
https://@               scheme=https  host=null  authority=[@]
https://user@/          scheme=https  host=null  authority=[user@]
https://:8443/          scheme=https  host=null  authority=[:8443]
https://user:pw@:8443/  scheme=https  host=null  authority=[user:pw@:8443]
```

All non-empty authorities, all pointing nowhere, all passing. My previous fix swapped "has a scheme" for "has an authority" and I treated authority as a proxy for hostname — which it isn't.

I didn't apply the suggested diff as written, because it adds a *second* notion of "has a host" beside the one `browsableHostOrNull` already uses. That's the same shape as the duplication Sourcery flagged earlier in this PR, and the consequence here is concrete: the predicate decides whether a URL has a destination and the label names that destination, so if they disagreed, one would cache a link the other couldn't describe.

Both functions now go through a single `hostnameOrNull()`. The rule is your rule — what survives removing userinfo and port must be non-empty and must not still start with the colon that introduces a port — but it lives in one place, so the predicate and the label cannot drift apart. IDN authorities still pass.

Two tests, both confirmed to fail if the `startsWith(':')` / non-empty check is relaxed back.

Three rounds on this helper now, each a case the previous fix didn't cover — scheme-only, then authority-as-host, now authority-without-host. That's the tell that I was validating against examples rather than against a definition, which is why this round replaced the check with a single derivation both callers share rather than adding another condition.

---

HeartlessVeteran2:
Real again — `https://host:abc/` has a non-numeric port, so the strip left it alone and `host:abc` passed both conditions. Fixed in `da5abe21`, but differently from the last three times, because that's now four rounds on twenty lines and the repetition is the finding.

**What I'd been doing wrong.** Each round I patched the shape that was named — `@`, then `:8443`, now `host:abc` — and each fix was locally correct. The shapes were never the problem: the fallback had no stated *reason* to exist, so there was nothing to check a new case against. Every review round was therefore guaranteed to find one more.

**The reason, once written down, is narrow.** `URI.getHost()` parses RFC 2396 and returns null for a non-ASCII name like `https://例え.テスト` — a URL browsers resolve fine. That is the only thing the fallback is for. Everything else `getHost()` refuses — `host_underscore`, `-lead.test`, `host:abc` — it refuses for a reason, and overruling a well-tested parser one example at a time is exactly how the previous rounds happened.

So the fallback now applies **only when the authority contains a non-ASCII character**, and only once it's free of the delimiters that mean it's carrying something other than a bare name. This is your "restrict the authority fallback to IDN hostnames" suggestion; I took the first of the two options rather than the leftover-colon one, because the colon rule would have been the fifth shape rather than a rule.

Probe confirming IPv6 never reaches the fallback, so brackets need no special handling:

```
https://[::1]:8080/   host=[::1]   authority=[[::1]:8080]
https://host:abc/     host=null    authority=[host:abc]
```

**The tests are now a table.** Authorities and whether each is reachable, with all four historical failures as rows alongside the valid cases — a new case is a row next to the others, which is where a gap in the reasoning is actually visible, rather than another method appended at the bottom. A second test walks the same table asserting the predicate and the label agree on every row, since one caching a link the other can't describe is precisely what sharing `hostnameOrNull` exists to prevent.

Both halves of the rule mutation-checked: dropping the non-ASCII requirement and dropping the delimiter check each fail both table tests.

---

HeartlessVeteran2:
Right again, and the suggested remedy is the one that ends this. `933b44cc`.

My round-4 fix asked whether a non-ASCII authority contained any of a handful of delimiter characters. That's a heuristic wearing a rule's clothing — it rejected the examples in front of it and said nothing about empty labels, leading or trailing hyphens, embedded spaces, over-long labels or emoji. All browsable chips.

`java.net.IDN.toASCII(name, IDN.USE_STD3_ASCII_RULES)` is the specification's own validator. Verified rather than assumed:

| candidate | `toASCII` | `toASCII` + STD3 |
|---|---|---|
| `例え.テスト` | `xn--r8jz45g.xn--zckzah` | same |
| `例え..テスト` | throws | throws |
| `例え.テスト-` | **accepts** | throws |
| `例え.テスト:abc` | **accepts** | throws |
| `例え テスト` | **accepts** | throws |
| `😀.test` | throws | throws |
| `host_underscore` | **accepts** | throws |

`USE_STD3_ASCII_RULES` is load-bearing — four of those pass without it. The conversion result is discarded on purpose: punycode is for resolvers, and the caller wants the name a user would recognise.

The non-ASCII gate stays in front, now earning its place twice. It still states *why* the fallback exists — `URI.getHost()` refuses non-ASCII names and nothing else needs rescuing — and it also blocks degenerate ASCII that IDN itself accepts: `IDN.toASCII(".")` returns `"."`, which the gate rejects and the validator wouldn't.

Both halves mutation-checked; dropping STD3 and dropping the gate each fail the grammar table. The malformed-Unicode cases are new rows in it.

---

Five rounds on this one file, so the honest summary: every fix before this one was "reject the character that was named". Each was locally correct and each guaranteed a successor, because I kept answering *which inputs are bad* instead of *what makes an input good*. The general form — when the same small file draws repeat findings, the missing thing is a definition, and the definition is usually a library that already implements the spec — is worth more than the fix. I'd written half of that lesson into my own notes after round 4 and still reached for a character class first.

---

Issue #1251, opened 2026-08-08
Feed tab is permanently empty: nothing has ever written a FeedItemEntity


HeartlessVeteran2:
Stage 6b. Investigated and verified against the code; this is a spec ready to execute, not a hypothesis.

## The Feed has never shown anything

`FeedRefreshWorker.doWork()` resolves enabled sources, purges items older than 30 days, and returns success. It never inserts anything. Its own KDoc admits it:

> *"new chapter items **would be** inserted by the source extension loader once implemented"*

Grepping `FeedItemEntity` across the repo: the entity, the DAO, one mapper and a test. `FeedDao.insertFeedItem` / `insertFeedItems` have **no production caller**. `FeedRepository` doesn't even expose a way to add one — the interface has `getFeedItems`, `refreshFeed`, `markFeedItemAsRead`, `clearFeedHistory`, and no writer.

So the tab renders an empty list forever, and the 6-hourly worker exists to garbage-collect rows that never arrive.

## Where the data already is

`LibraryUpdateWorker` line ~136 already has exactly what the feed needs:

```kotlin
val result = updateLibraryManga(manga)
result.onSuccess { newChapterCount -> … }
```

and `UpdateLibraryMangaUseCase` already computes the new chapters — it diffs source chapters against stored ones by URL, inserts them, and then **throws them away**, returning `newChapters.size`.

## The work

1. **`UpdateLibraryMangaUseCase` returns the chapters, not a count** (`Result<List<Chapter>>`). Note the wrinkle: the chapters it builds have `id = 0` because Room assigns ids on insert, and `FeedItemEntity.chapterId` needs the real one. So after `insertChapters`, re-read by manga id and filter to the new URLs. One extra read, only when there are new chapters.
2. **`FeedRepository.addFeedItems(items)`** + impl over `feedDao.insertFeedItems`.
3. **`LibraryUpdateWorker` writes feed items** in the existing `onSuccess` branch. `sourceName` comes from `sourceRepository.getSource(manga.sourceId.toString())?.name`; `newChapterCount` becomes `newChapters.size` at the two sites that use it.
4. **~30 test call sites** in `LibraryUpdateWorkerTest` use `Result.success(<Int>)`. Mechanical: a `newChapters(n)` helper and a regex replace.

## Separately: `addSource` hashes text the user typed

`SavedFeedViewModel.addSource()`:

```kotlin
[...]

PR #1253, opened 2026-08-09
Make the Feed show something

HeartlessVeteran2:
Closes the writer half of #1251. Stage 6b.

## The Feed tab has never displayed anything

`FeedRefreshWorker.doWork()` resolves enabled sources, purges items older than thirty days, returns success. It never inserts. Its own KDoc admits it:

> *"new chapter items **would be** inserted by the source extension loader once implemented"*

`FeedDao.insertFeedItem` / `insertFeedItems` had **no production caller**, and `FeedRepository` didn't even expose a writer — reads, a refresh trigger, mark-as-read, clear, and no way to add one. So a 6-hourly worker has been garbage-collecting rows that never arrive.

## The data was already in hand

`UpdateLibraryMangaUseCase` diffs source chapters against stored ones by URL, inserts the new ones — and then throws them away to return `newChapters.size`. It returns the chapters now, which is what a caller needs to *name* what arrived rather than count it.

One wrinkle worth stating, because it's the kind of thing that ships broken: the chapters it builds carry `id = 0`, since Room assigns the id on insert. A feed row that can't open its chapter isn't much of a row, so the use case re-reads after inserting and filters to the new URLs. One extra read, and only when something was actually new.

## Two judgement calls in the worker

- **The source *name* is looked up, not derived from the id**, because the feed shows it to the user. An extension can be uninstalled while its manga stay in the library, so a missing source falls back to the id rather than dropping the row — the chapter genuinely arrived, and hiding it because its extension went away is a worse answer than an unfamiliar label.
- **`dateUpload == 0` is stamped with now.** The feed orders by timestamp, so an undated chapter reported as 0 would be pinned to 1970 and buried under everything else.

The write is guarded by `runCatching`. A feed row is a nicety; the auto-download pass immediately after it is not. Failing the whole library update because a secondary write failed would trade a real feature for a cosmetic one.

## Verification

The ~23 existing worker tests moved from `Result.success(n)` to a `newChapters(n)` helper — mechanical, since they only ever cared how many.

Five new tests, all mutation-checked with `--rerun-tasks`:

| Mutation | Tests that failed |
|---|---|
| Remove the feed write | `new chapters are recorded in the feed`, `a chapter from an uninstalled source still reaches the feed`, `an undated chapter is stamped now rather than 1970` |
| Let a feed failure propagate | `a feed write failure does not fail the library update` |

## Not here

`SavedFeedViewModel.addSource()` hashes user-typed text into a source id that can never match a real source (#1251). It looks like it would block this, and it doesn't: `FeedViewModel` reads `getFeedItems(limit)`, **unfiltered by feed source**, so the `feed_sources` table never gated what the tab shows. Fixing the picker alone would have changed nothing visible; this is the half that makes the feature exist.

---

## Summary by Sourcery

Populate the Feed screen with newly arrived chapters from library updates and expose the necessary write path so feed items are stored.

New Features:
- Add a write API to FeedRepository and implement persistence of new feed items so the Feed tab can display recent chapters.
- Make LibraryUpdateWorker convert newly fetched chapters into feed items with source and manga metadata for display in the Feed screen.

Enhancements:
- Change UpdateLibraryMangaUseCase to return the list of newly inserted chapters instead of just a count, enabling downstream features like the feed to reference specific chapters.
- Ensure undated chapters receive a current timestamp and that feed writes are best-effort so library updates and auto-downloads are not blocked by feed failures.

Documentation:
- Clarify CLAUDE documentation for the Feed screen to describe its behavior and the new write path from library updates.

Tests:
- Extend LibraryUpdateWorker tests to cover feed item creation, handling of uninstalled sources, timestamping of undated chapters, and resilience to feed write failures.
- Update UpdateLibraryMangaUseCase tests to validate the new list-based result contract and behavior when no new chapters or an empty source result are found.

---
## Summary by cubic
Makes the Feed tab show new chapters by writing feed items during library updates, and makes each feed row unique per chapter to prevent duplicates. Also fixes cancellation handling so a cancelled update stops cleanly, and reattaches two misplaced KDocs.

- **New Features**
  - Added `FeedRepository.addFeedItems` and implemented it in `FeedRepositoryImpl`.
  - `LibraryUpdateWorker` writes feed items for new chapters; looks up the source name (falls back to id), stamps `dateUpload == 0` with now, and guards writes so failures don’t fail the update.

- **Migration**
  - Database v45: add a unique `(mangaId, chapterId)` index for `feed_items`; DAO inserts use `IGNORE`; migration deduplicates existing rows.
  - `UpdateLibraryMangaUseCase` now returns `Result<List<Chapter>>` instead of `Result<Int>`; update callers (e.g., `result.getOrNull()?.size ?: 0` for counts).

<sup>Written for commit 48c3def945481bc0f4dd3cdcef3cddbf4a30c8df. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/HeartlessVeteran2/Otaku-Reader/pull/1253?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


HeartlessVeteran2:
All five real. Four fixed in `0bbfa673`; the fifth is app-wide and filed as #1254.

**Duplicate feed rows (cubic P2 / CodeAnt) — correct.** Verified the premise rather than assuming it: `WORK_NAME = "library_update"` and `PERIODIC_WORK_NAME = "library_update_periodic"` really are different unique-work names, so a manual and a periodic update can overlap, and `feed_items` had no uniqueness constraint.

What settled the fix was noticing that **`chapters` already defends against this exact race** with a unique `(mangaId, url)` index. The race isn't hypothetical — the schema conceded it long ago, and the feed simply wasn't given the same treatment. So `(mangaId, chapterId)` is unique now (schema v45). I chose that over serializing the workers because a database constraint can't be undone by a timing change somewhere else, and serializing is a user-facing behaviour change besides.

One consequence worth flagging: the insert strategy had to move to **IGNORE**, not REPLACE. With a unique index, REPLACE deletes the conflicting row and inserts a new one — fresh id, `isRead` reset — so a feed item the user had already read would pop back as unread every time an overlapping update re-recorded it.

**`runCatching` swallowing `CancellationException` (cubic P2) — correct**, and it's the rule this codebase applies at every other catch site. Narrowed to `catch (e: Exception)` with cancellation rethrown.

Worth reporting honestly: my first mutation check on this found **no test failing** when I put the swallow back. The guard was unexercised. I added a test before shipping it rather than leaving a fix that nothing holds — that's the "prose is where a claim goes when no test holds it" rule biting on my own code.

**Both P3 test findings — correct, and both were the "assertion that would pass with the thing it names deleted" shape.** The use-case test claimed to verify the new chapters and asserted only `.size`, which is precisely the count assertion it replaced; it asserts the URLs now, and dropping chapter identity from the return makes it fail. The timestamp test bounded "stamped now" from below only, so any non-zero value — including a far-future one — passed; bounded both ways now.

---

**The fifth finding is real and bigger than this PR** (cubic on `UpdateLibraryMangaUseCase:72`): overlapping updates can leave a feed row pointing at a `chapterId` the other run's REPLACE reassigned.

I traced it, and it isn't a feed problem. `chapters` inserts with `OnConflictStrategy.REPLACE` on a unique index, which deletes and re-inserts with a **new autogenerated id** — so chapter ids are not stable, and `download_queue`, `reading_history`, `page_bookmarks`, `reader_comments` and `sync_queue` all store one. A history entry that won't resume or a bookmark that opens nothing are the same bug.

Fixing it means changing the conflict strategy on every chapter write path, which is not something to do inside a Feed change. Filed as **#1254** with the mechanism, the affected tables, and the trade-off (IGNORE preserves ids but stops genuine metadata refreshes applying, so it probably wants a real upsert).

Full verification green; every new guard mutation-checked with `--rerun-tasks`.

---

HeartlessVeteran2:
Both correct, fixed in `48c3def9`.

`FeedItemEntity` genuinely had two consecutive doc blocks — I prefixed the new one to the `@Entity` annotation without noticing the file already carried `/** Entity for feed items - latest chapters from various sources. */` above it. Kotlin binds the nearest preceding KDoc, so the original became dead text sitting in the source, which is worse than no comment because it still reads like documentation. Merged into one block that keeps the original's summary line.

`timestampToleranceSeconds` was inserted between `newChapters`' KDoc and `newChapters`, so the doc described the constant. Moved up beside the other test fields with its own one-liner.

Marked P3, and cosmetically they are — but the mechanism isn't cosmetic and it's the third time I've done it on this project (there was a stranded `METADATA_QUERY` KDoc in #1249). The cause is always the same: an anchored text insertion that lands between a doc block and its declaration, which compiles fine and reads fine in the diff because the diff shows the added lines, not what they separated. The check that would catch it is looking at the *result* around an insertion point rather than the patch — worth writing down since noticing it three times hasn't stopped it happening.

Full verification green.

---

Issue #1256, opened 2026-08-09
Download folders are named by the numeric source key; switching to display names needs a migration


HeartlessVeteran2:
Split out of the source-key round-trip fix, because it is the one part that cannot be an edit.

## What it does today

`resolveDownloadFolderName(sourceId: Long)` used to read:

```kotlin
getSource(sourceId.toString())?.name ?: sourceId.toString()
```

and its KDoc claimed it returned "the source's actual display name (e.g. "MangaDex") when the source can be looked up". It never could. `sourceId` on a manga row is `MangaSource.id.hashCode().toLong()`, so `sourceId.toString()` is a hash's decimal and matches no source's id. The lookup missed every time and the fallback ran every time.

So **every download on every device is filed under the number**, and has been since the feature shipped. The function has been simplified to say so honestly:

```kotlin
suspend fun SourceRepository.resolveDownloadFolderName(sourceId: Long): String = sourceId.toString()
```

## Why it wasn't just "fixed"

The broken lookup is load-bearing. Repairing it so the name resolves would point every download *read* at `Downloads/MangaDex/…` while every existing file sits in `Downloads/-1874553621/…`. Chapters already on disk would silently vanish from the app — the worst possible outcome for offline reading.

A test now pins the absence of the lookup (`resolveDownloadFolderName is the numeric key and consults no source`) so nobody "fixes" it back by accident.

## What a real change looks like

1. Resolve each installed source's canonical key → display name.
2. Rename `<downloadRoot>/<key>` → `<downloadRoot>/<name>` on disk, for keys that resolve.
3. Leave unresolvable keys (uninstalled extensions) under their number — renaming those has no name to rename to, and they must stay readable.
4. Handle the collision case: two sources whose display names are equal, and a target directory that already exists.
5. Make it idempotent and resumable — it is filesystem work that can be interrupted mid-way, and a half-renamed download root must not lose chapters.
6. Only then change `resolveDownloadFolderName`, and update its KDoc and test together.

Worth doing — folder names a human can read are a genuine improvement, and it also makes backups portable between installs — but it is a data migration with a real failure mode, not a one-line change.

PR #1257, opened 2026-08-09
Fix the source-key round-trip so a library manga can reach its source

HeartlessVeteran2:
Found while starting the feed source picker (#1251) — the picker hashes user-typed text into a source id, and chasing *why* that was wrong led here.

## The bug, in one line

A manga row stores `MangaSource.id.hashCode().toLong()`. Every read-back does `manga.sourceId.toString()`. **Hashing is one-way** — that string is the decimal of a hash, and it matches no source's id.

## What that means in practice

`SourceRepositoryImpl.getSource` is `_sources.value.find { it.id == sourceId }`, so the lookup misses, and `getPageList` / `getChapterList` return `Result.failure(IllegalArgumentException("Source not found: …"))`. That call pattern was in:

| Where | Symptom |
|---|---|
| `UpdateLibraryMangaUseCase` | library update finds no chapters, ever |
| `ReaderChapterLoaderDelegate` | reading any chapter that isn't already downloaded fails |
| `RecommendationRepositoryImpl` | candidate pool is always empty |
| `MigrateMangaUseCase` / `SearchMigrationTargetsUseCase` | migration can't reach the target source |
| `DetailsViewModel` ×7 | blank source name, no page thumbnails, no suggestions, broken tag→search |
| `AdaptiveChapterPrefetcher`, `ReaderDownloadAheadDelegate` | prefetch and download-ahead silently no-op |

Verified rather than assumed — a five-line `javac` probe over the three id shapes the app actually produces:

```
2499283573021220255 -> key -1874553621 -> reverse "-1874553621" matches id? false   (APK extension)
local               -> key   103145323 -> reverse "103145323"   matches id? false   (LocalSource)
en.mangadex         -> key  1639958408 -> reverse "1639958408"  matches id? false   (JS source)
```

**Browse was never affected**, which is why the app looked half-working: `Route.SourceListing.sourceId` is already a `String`, so browsing carries the real id and never goes through the key. Everything reached *from the library* went through it.

## The fix

Hashing can't be inverted, so the reverse has to be a search:

```kotlin
suspend fun getSourceByKey(key: Long): MangaSource?        // on SourceRepository
suspend fun SourceRepository.resolveSourceId(key: Long): String?
```

Every call site converted.

### Why `getSourceByKey` tries two rules

There were **three** conventions in the code for turning a source id into a `Long`, all live at once:

- `toSourceId()` — hash. `AddMangaToLibraryUseCase`, the main add-to-library path.
- `toLongOrNull()` — parse. `SourceMangaDetailViewModel`, reached by tapping a search result.
- the raw `ExtensionSource.id`. The library's icon map.

So whether a manga worked depended on which screen added it. Going forward everything writes the canonical `toSourceId()` key, but rows already on disk are just `Long`s — nothing records which rule wrote them, so a migration cannot separate them. `getSourceByKey` therefore tries the canonical rule and falls back to the parse, canonical first so a legacy match can't shadow a source that genuinely owns the key today. There's a test for exactly that collision.

## Four more instances of the same confusion, fixed here

- **Library source name and language** were keyed by `source.id.toLongOrNull()`, so `sourceName` was `""` for every entry — which is why "group by source" showed numbers.
- **Library source icons** were keyed by the raw `ExtensionSource.id`; the canonical key is that id *stringified and hashed* (`TachiyomiSourceAdapter.id` is `tachiyomiSource.id.toString()`). No entry ever matched, so no icons.
- **Migration** mapped every non-numeric source (`local`, every JS source) onto `0L`, so they collided on one `LazyColumn` key.
- **`isManhwa`** did `manga.sourceId.toString().lowercase().contains("webtoon")` — searching a decimal number for a word. The Manhwa tab was permanently empty.
- **`DownloadManager.resolvePageUrls`** passed `request.sourceName` — the download *folder* name — as a source id, so any download enqueued without page URLs resolved to an empty list and went straight to `FAILED`.

## What is deliberately *not* fixed

`resolveDownloadFolderName` used to be `getSource(sourceId.toString())?.name ?: sourceId.toString()`, with a KDoc promising the display name. It never resolved, so **every download on every device is filed under the number**.

Making it resolve now would point every download *read* at `Downloads/MangaDex/…` while the files sit in `Downloads/-1874553621/…` — chapters already on disk would vanish from the app. The broken lookup was load-bearing. It's now `= sourceId.toString()`, which is what it always did, with a KDoc that says so and a test asserting it consults *no* source, so nobody "fixes" it back by accident. Doing it properly is a filesystem migration: **#1256**.

The same trap bit `ReaderChapterLoaderDelegate`, where one variable was feeding both `getPageList` (wants a source id) and `PageLoader.resolveUrl` (wants a folder name). They were the same string only because resolution was broken; they're now two variables, so fixing #1256 later can't silently break the reader.

## Tests

New:
- `getSourceByKey` — finds the source that produced the key, for numeric and non-numeric ids; matches legacy parsed rows; prefers the canonical rule on collision; null when nothing owns the key. The first of those also asserts `getSource(key.toString())` **misses**, or it would prove nothing.
- `UpdateLibraryMangaUseCase` — asks by the resolved id, explicitly `coVerify(exactly = 0)` on the stringified key; fails when nothing owns the key.
- `SearchMigrationTargetsUseCase` — fails when the target key doesn't resolve.
- `resolveDownloadFolderName is the numeric key and consults no source`.

Rewritten, and worth flagging: `SourceRepositoryExtTest` previously asserted `resolveDownloadFolderName` returns the display name — and passed, because it stubbed `getSource("1943584017")` to return a source. Nothing in production ever registers a source under that id. A green test asserting behaviour the app has never had.

Each new guard was checked by mutation with `--rerun-tasks`: reverting the resolution, deleting the legacy pass, swapping the rule precedence, and making the folder name resolve again each fail exactly the test that names them.

`CLAUDE.md` gains a *Source Identity* section under Architecture and a new Common Bug Areas entry.

---

## Summary by Sourcery

Ensure library and reader flows resolve sources from the hashed Long keys stored on manga rows instead of stringified hashes, and document the two-step source identity model.

Bug Fixes:
- Fix source lookups that passed manga.sourceId.toString() by resolving the real string source id via getSourceByKey/resolveSourceId, restoring library updates, reading, migration, recommendations, and suggestions for library manga.
- Correct download page URL resolution to derive the source from the manga’s key rather than the download folder name, so downloads enqueued without page URLs can be resolved.
- Keep download folder names as the numeric key string to avoid orphaning existing on-disk downloads while clarifying this behavior in tests and docs.
- Fix library source name, language, icon, migration target ids, and manhwa detection by consistently keying on the canonical hashed source key instead of mixed raw/parsed ids, so grouping and UI filters behave correctly.

Enhancements:
- Add getSourceByKey and resolveSourceId helpers to bridge between hashed Long keys on manga rows and string source ids, including support for legacy rows written with numeric ids.
- Refine prefetching, download-ahead, migration, and worker logic to use resolved source ids and clearer separation between source ids and download folder names.
- Expand architecture docs with a Source Identity section and update common bug guidance around source-related issues.

Tests:
- Add comprehensive tests for getSourceByKey and resolveSourceId behavior, including legacy numeric rows and collision precedence.
- Strengthen tests for update-library, migration target search, download folder naming, reader behavior, library icons, and library updates to assert correct source resolution and to guard against regressions.

---
## Summary by cubic
Fixes the source key round-trip so library manga can reach their source again, centralizes the key rule, and restores reading, updates, migration, recommendations, prefetch, and download-ahead. Also stops recommending manga you already own while keeping download folders stable.

- **Bug Fixes**
  - Added `SourceRepository.getSourceByKey` and `resolveSourceId`; all call sites now resolve the real source id instead of passing a stringified hash.
  - Centralized key indexing in `associateBySourceKey`; `getSourceByKey` now delegates to it for one rule and consistent collision handling (canonical hash first, with legacy parse fallback).
  - Recommendations seed only from resolved sources, apply the cap after resolving, persist the canonical key, and exclude manga already in the user’s library.
  - `DownloadManager` resolves the source from the manga via `MangaDao` and uses the resolved id for page lists; downloads from an unresolvable source now move to FAILED instead of stalling; the folder name is never used as a source id.
  - `Reader` returns `Result.NotFound` when a source can’t be resolved and keeps source id separate from the download folder name.
  - Library source names, icons, migration lists, filters, and Manhwa tab detection now key off the canonical source key, fixing empty values and wrong grouping across the app.
  - Tests cover key reversal, collision precedence, recommendation seeding (including owned-manga exclusion), and assert the folder name is never used as a source id.

- **Migration**
  - No data migration. `downloadFolderNameFor` always returns the numeric key string to keep existing downloads visible; switching to display names is tracked in #1256.

<sup>Written for commit d82d59e5d87c3ba2513f6b0d1d191fb88e958c4a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/HeartlessVeteran2/Otaku-Reader/pull/1257?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -187,6 +187,36 @@ LibraryViewModel.kt — @HiltViewModel, produces StateFlow<LibraryState>
 LibraryScreen.kt    — stateless composable consuming state
 ```
 
+### Source Identity — Two Ids, One Direction
+
+A source has a **string id** (`MangaSource.id`) and every manga row stores a **`Long` key**
+(`Manga.sourceId`). The key is `id.toSourceId()`, which is `id.hashCode().toLong()`.
+
+**Hashing is one-way.** `manga.sourceId.toString()` is the decimal of a hash and matches no
+source's id — it is *not* a way back. To go from a key to a source, search the loaded sources:
+
+```kotlin
+sourceRepository.getSourceByKey(manga.sourceId)          // the MangaSource
+sourceRepository.resolveSourceId(manga.sourceId)         // its string id, for a source call
+```
+
+Getting this wrong is the highest-impact bug the project has had. `getSource(manga.sourceId.toString())`
+was in reading, library update, recommendations, migration, the details screen, prefetch and
+download-ahead; all of them failed with "Source not found" for every library entry, and the
+details header's source name was permanently blank. Three *different* wrong conventions were in
+use at once (`toSourceId()`, `toLongOrNull()`, and the raw `ExtensionSource.id`), which is why
+some manga worked and others didn't depending on which screen added them. Fixed across the repo;
+`getSourceByKey` also matches the legacy `toLongOrNull()` rows, because a `Long` on disk carries
+no record of which convention wrote it and so no migration can tell them apart.
+
+Two things are deliberately *not* this key:
+
+- **`resolveDownloadFolderName`** returns the numeric key as a string and consults no source. Every
+  download on disk is already filed under the number, so resolving a display name would orphan
+  them. Changing it is a data migration — see #1256.
+- **`Route.SourceListing.sourceId`** is already the string id. Browse never went through the key,
+  which is why browsing worked while reading from the library did not.
+
 ### Clean Architecture Layer Rules
 
 | Layer | Contains | Rules |
@@ -472,12 +502,13 @@ A claim about a library's published versions was taken from a search API that on
 2. **Room DAO not connected** — DAO not injected into repo, repo not injected into UseCase. Trace the chain.
 3. **MVI state not updating UI** — `StateFlow` not collected in Compose, or reducer emitting same reference. Use `copy()`.
 4. **Extension loader failures** — ClassLoader issue, missing permission, or interface mismatch with Tachiyomi API.
-5. **Gradle dependency conflicts** — Version mismatches between Compose BOM, Kotlin, Hilt, or KSP. Check `libs.versions.toml` first.
-6. **Navigation crashes** — Missing destination, wrong argument type in NavGraph, or missing `@Serializable` on route class.
-7. **Coroutine scope leaks** — `GlobalScope` used instead of `viewModelScope` or `lifecycleScope`. Always use structured concurrency.
-8. **Stale undo from concurrent batches** — Guard the undo handler: `if (pendingBatchIds != incomingIds) return`.
-9. **Flow recreated on recomposition** — Wrap with `remember(key) { flow }` when derived from a `@Singleton` injected dependency.
-10. **Room migration FK violations** — Recreating a table (to DROP a column) while child tables have FK references to it causes `SQLITE_CONSTRAINT_FOREIGNKEY`. Wrap the CREATE/INSERT/DROP/RENAME block with `PRAGMA foreign_keys = OFF` before and `PRAGMA foreign_keys = ON` after.
+5. **"Source not found" for a library manga** — someone passed `manga.sourceId.toString()` where a source id was wanted. Use `getSourceByKey` / `resolveSourceId`. See *Source Identity* above.
+6. **Gradle dependency conflicts** — Version mismatches between Compose BOM, Kotlin, Hilt, or KSP. Check `libs.versions.toml` first.
+7. **Navigation crashes** — Missing destination, wrong argument type in NavGraph, or missing `@Serializable` on route class.
+8. **Coroutine scope leaks** — `GlobalScope` used instead of `viewModelScope` or `lifecycleScope`. Always use structured concurrency.
+9. **Stale undo from concurrent batches** — Guard the undo handler: `if (pendingBatchIds != incomingIds) return`.
+10. **Flow recreated on recomposition** — Wrap with `remember(key) { flow }` when derived from a `@Singleton` injected dependency.
+11. **Room migration FK violations** — Recreating a table (to DROP a column) while child tables have FK references to it causes `SQLITE_CONSTRAINT_FOREIGNKEY`. Wrap the CREATE/INSERT/DROP/RENAME block with `PRAGMA foreign_keys = OFF` before and `PRAGMA foreign_keys = ON` after.
 
 ---
 
--- a/data/src/main/java/app/otakureader/data/download/DownloadManager.kt
+++ b/data/src/main/java/app/otakureader/data/download/DownloadManager.kt
@@ -5,6 +5,7 @@ import android.util.Log
 import app.otakureader.core.common.network.NetworkMonitor
 import app.otakureader.core.common.network.NetworkType
 import app.otakureader.core.database.dao.DownloadQueueDao
+import app.otakureader.core.database.dao.MangaDao
 import app.otakureader.core.database.entity.DownloadQueueEntity
 import app.otakureader.core.preferences.CbzEncryptionStore
 import app.otakureader.core.preferences.DownloadPreferences
@@ -13,7 +14,6 @@ import app.otakureader.domain.model.DownloadItem
 import app.otakureader.domain.model.DownloadPriority
 import app.otakureader.domain.model.DownloadStatus
 import app.otakureader.domain.repository.ChapterRepository
-import app.otakureader.domain.repository.MangaRepository
 import app.otakureader.domain.repository.SourceRepository
 import app.otakureader.domain.repository.resolveSourceId
 import app.otakureader.sourceapi.SourceChapter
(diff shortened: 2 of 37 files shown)


HeartlessVeteran2:
Pushed `d4dc1a67`. Rundown of what landed and what didn't.

## Accepted

**The duplicated key rule (cubic ×3, CodeAnt ×1).** This was the important one and it was right. `getSourceByKey` matches both the canonical hashed key and the legacy parsed one, but I then hand-wrote three map builders — library source names, library icons, merge-duplicates — that indexed only the canonical key. Same rule, four implementations, one of them narrower: a legacy row resolved through `getSourceByKey` but came back blank in the library.

Rather than adding the fallback in three places, the rule now lives once as `associateBySourceKey`, including the collision precedence (legacy written first so the canonical key overwrites it, matching `getSourceByKey`). Four tests, two mutation-checked: dropping the legacy pass and swapping the precedence each fail the test that names them.

**`resolveDownloadFolderName` shouldn't be a `SourceRepository` extension (Sourcery).** Agreed — it consulted no source, and the `@Suppress("UnusedReceiverParameter")` was the tell. It's now the top-level `downloadFolderNameFor`. That paid off immediately: detekt then flagged **six** classes injecting `SourceRepository` solely to call it, all now dropped.

**The `DownloadManager` cycle (Sourcery).** Also agreed, and fixable rather than documentable — it now takes `MangaDao`, which has no edges back, matching the `DownloadQueueDao` this class already injects. No cycle to rely on a `Lazy` for.

**Resolve before insert in `MigrateMangaUseCase` (cubic).** Correct and I'd missed it — an unresolvable target failed *after* `insertManga`, leaving a library row for a source nothing can reach. Resolution moved above the insert.

**Recommendation seeding (CodeAnt + cubic).** Both right, and they were different bugs: the cap was applied before resolving (so uninstalled sources burned slots while installed ones further down were never reached), and new candidate rows copied the library row's key, which could be a legacy one — writing the same source under two keys. Cap now applies after resolving; rows persist the canonical key.

**The `checkNotNull` in the reader (Sourcery).** The premise was off — `load()` wraps the call in `try/catch` returning `Result.Failure`, so it was never a crash — but the surrounding code makes a better case than either version: `Chapter not found` and `Manga not found` are already `Result.NotFound` two lines above, and an uninstalled extension belongs with them. Resolution moved up beside its siblings and the throw is gone.

**The download-resolution test (cubic).** The stated outcome was wrong — the test passes — but the substance was right, and worse than described. `mangaDao`/`sourceRepository` were relaxed mocks, and MockK relaxed mode returns a *child mock* for a nullable return, not null, so the source resolved to one with `id == ""` and `getPageList(any(), any())` matched it. Green while proving nothing. It now stubs a real key→source resolution, asserts the resolved id reaches `getPageList`, asserts the folder name never does, and there's a second test for the unresolvable case. Mutation-checked: restoring `request.sourceName` as the source id fails it.

## Not accepted

**`ReaderViewModelTest`: "MockK returns `null` for a nullable return, so the comment is wrong" (CodeAnt).** The comment is accurate. Probed it directly rather than argue from memory — a relaxed `SourceRepository`, unstubbed:

```
PROBE relaxed getSourceByKey(1L) = MangaSource(child of #1#2)
PROBE id =
```

A child mock with an empty `id`, not null. That behaviour is exactly what made the download test above pass for the wrong reason, so the comment is load-bearing and stays.

## Split out

**Startup race on `refreshSources()` (CodeAnt).** Real, but pre-existing — `getSource(String)` has always read the same unguarded snapshot; the round-trip fix just makes the window visible by making those paths capable of succeeding. The obvious fix (await a non-empty `_sources`) would hang forever for a user with no extensions installed, so it needs an explicit initial-load signal and some decisions about failure and timeout behaviour. That changes what a source lookup promises, so it's **#1258** rather than a rider on a bug fix.

---

HeartlessVeteran2:
Pushed `d82d59e5`. All three taken — and the P3 was the best finding of the round.

**The two implementations of one rule (P3).** I claimed last round that the key rule "now lives once." It didn't. `associateBySourceKey` held it, and `getSourceByKey` still spelled it out separately — and they disagreed exactly where it's hardest to notice: a linear `find` keeps the *first* match in list order, a map built with `put` keeps the *last*. Same key, two sources, different answers depending on which path asked. That's my own most common defect: prose asserting a property the code has on only one path.

`getSourceByKey` is now `_sources.value.associateBySourceKey { it.id }[key]`. One implementation, so the tie-break agrees by construction rather than by maintenance.

Worth noting the collision is not hypothetical — `"Aa"` and `"BB"` both hash to 2112, so the new test uses those rather than a contrived stub, and asserts not that a particular source wins but that **both paths return the same one**. Reverting `getSourceByKey` to its own `find` fails it.

**Recommending the user their own manga (P2).** Correct, and a regression I introduced *in the previous commit* — the fix for your earlier canonical-key finding caused it. Candidates are written under the canonical key; a legacy library row holds the parsed one; they don't collide on the `(sourceId, url)` unique index, so `insertIfNotExists` adds a second non-favourite copy. Nothing downstream catches it either, because the library exclusion filters by row id and the new row has a different one.

Both keys stay correct — the missing piece was deduping through the *resolved source* rather than the raw key, which recognises the row under either convention. New `RecommendationRepositoryImplTest` covers it, plus the seed-budget case from your earlier finding.

**The FAILED assertion (P2).** Right, and it's the "assert what it left behind, not just what didn't happen" rule. The test only checked absent calls, so a regression parking the download in QUEUED forever would have passed — precisely what it exists to catch. Asserting the terminal state now.

Every new guard mutation-checked with `--rerun-tasks`: dropping the owned-URL filter, moving the cap back before resolution, and un-sharing the key rule each fail the test that names them.

---

PR #1259, opened 2026-08-09
Pick feed sources from the installed list instead of typing a name

### LINKED ISSUE #1251: Feed tab is permanently empty: nothing has ever written a FeedItemEntity

HeartlessVeteran2:
## **User description**
Part 2 of #1251. The writer landed in #1253 (the Feed tab now has content); this is the other half, which #1251 flagged as independent — and it is: nothing in the tab changes.

## What was wrong

`SavedFeedViewModel.addSource` took whatever the user typed and hashed it:

```kotlin
val sourceId = sourceName.hashCode().toLong().let { if (it < 0) -it else it }
feedRepository.addFeedSource(sourceId, sourceName)
```

It never consulted `SourceRepository`. A real source's key is `MangaSource.id.toSourceId()` — a hash of its *id*, not of its display name — so `"MangaDex".hashCode()` matches nothing, and neither would any other string. Every row `feed_sources` has ever held is unusable, and a typo was indistinguishable from a correct entry because neither was ever checked against anything.

## What changed

`FeedBuilderBottomSheet` is a searchable list of installed sources rather than a text field. `SavedFeedViewModel` combines `getFeedSources()` with `getSources()`, so what's on offer is "installed, minus what you already added", and keys come from `id.toSourceId()` — the canonical convention settled in #1257.

They're combined rather than collected separately because the available list is *defined* by both: a change to either side has to recompute both halves together.

## The filter is load-bearing

Excluding sources already in the feed looks cosmetic. It isn't:

```kotlin
@Insert(onConflict = OnConflictStrategy.REPLACE)
suspend fun insertFeedSource(source: FeedSourceEntity): Long
```

`feed_sources` has a unique index on `sourceId` and an autogenerated primary key, so REPLACE on a conflict **deletes the existing row and inserts a new one** — new `id`, and `isEnabled` / `itemCount` / `order` all back to defaults. Re-adding a source the user had disabled would silently re-enable it and drop it to the bottom of the order. Filtering it out of the picker is what makes that unreachable, so there's a test naming that reason.

I left the DAO alone rather than switching it to `IGNORE`: `insertFeedSources` (plural) is the backup-restore path, where REPLACE is the semantics you want — the restored row should win.

## Nothing visible changes in the Feed tab

Worth being explicit, since "fix the feed source picker" sounds like it should. `FeedViewModel` reads `feedRepository.getFeedItems(limit)`, which is unfiltered by feed source, so `feed_sources` does not gate what the tab displays. The value here is that the rows stop being junk — which is what any future per-source filtering, or the saved-searches feature that shares the table, would need.

## Tests

New `SavedFeedViewModelTest`:
- Sources are offered under the canonical key, and explicitly **not** `name.hashCode()` — asserting the two differ, or the first assertion would prove nothing.
- A source already in the feed is not offered again; removing it puts it back on offer.
- Adding stores the key the picker offered.
- A failed add surfaces a snackbar rather than throwing.

Mutation-checked with `--rerun-tasks`: dropping the already-added filter fails two tests, and going back to hashing the name fails four.

Also added `implementation(projects.sourceApi)` to `feature/feed` — `source-api` is not exposed transitively through `:domain`, matching how `feature/browse`, `feature/library` and `feature/migration` each declare it.

Closes #1251.

---

## Summary by Sourcery

Replace the feed source text input with a picker of installed sources keyed by canonical source IDs and keep the feed sources list consistent with installed sources.

New Features:
- Add a searchable bottom sheet listing installed sources for adding to the feed, including language labels and empty-state messaging.

Bug Fixes:
- Ensure added feed sources use the canonical MangaSource.id-based sourceId instead of hashing arbitrary user input, so saved feed sources correspond to real sources.
- Prevent re-adding an existing feed source from resetting its enabled state, item count, and ordering by filtering already-added sources out of the picker.

Enhancements:
- Combine feed source and installed source streams in SavedFeedViewModel to derive both current feed sources and addable sources in a single state.
- Update SavedFeed UI and state to work with structured FeedSourceOption objects instead of free-text names.
- Document feed source picker behavior and feed-screen semantics in CLAUDE.md.

Build:
- Add source-api as a direct dependency of the feed feature module.

Tests:
- Add SavedFeedViewModel tests covering canonical key usage, filtering of already-added sources, reinsertion behavior, correct key persistence on add, and snackbar behavior on add failures.

___

## **CodeAnt-AI Description**
Let users add valid installed sources to their feed

### What Changed
- Replaced free-text source entry with a searchable list of installed sources.
- Uses each source’s valid identifier, preventing added feed sources from becoming unusable because of typos or name-based IDs.
- Hides sources already in the feed to prevent resetting their enabled state, item count, or order.
- Shows an explanation when all installed sources have already been added.
- Displays an error message when adding a source fails.

### Impact
`✅ Usable feed source entries`
`✅ No duplicate additions that reset source settings`
`✅ Searchable installed-source picker`

### Talking to CodeAnt AI
<pre>
</pre>

#### Example
<pre>
</pre>

<pre>
</pre>

#### Example
<pre>
</pre>

### Retrigger review
<pre>
</pre>

</details>

---
## Summary by cubic
Replace the free-text feed source input with a searchable picker of installed sources, saving canonical keys via `id.toSourceId()` and blocking duplicates. Also flag saved sources that no installed source owns and disable their toggle; the Feed tab display remains unchanged.

- **New Features**
  - Picker in `FeedBuilderBottomSheet` lists installed sources with search and add-by-tap.
  - `SavedFeedViewModel` offers “installed minus already added”.

- **Bug Fixes**
  - Use real source IDs from `id.toSourceId()` instead of hashing typed names.
  - Filter out already-added sources to avoid `REPLACE` resetting enabled, count, and order.
  - Show “not installed” for rows whose key matches no installed source, keep the real source addable, and disable that row’s switch.

<sup>Written for commit afa886157e0ea7072ca01502f5775668383f5b23. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/HeartlessVeteran2/Otaku-Reader/pull/1259?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


HeartlessVeteran2:
Pushed `afa88615`. Taken, though not as suggested — the suggested fix has a worse failure mode than the bug.

**The finding is right.** A row written by the old sheet holds `"MangaDex".hashCode()`, the picker's filter compares against `id.toSourceId()`, and those differ — so the source is offered again and the user ends up with two rows for it.

**Why not normalise or dedupe by name.** Both proposed remedies mean matching a stored display name against an installed source's name. That is the same guessing that produced these rows: the name was free text, it need not match exactly, and two sources can share a display name across languages. Worse, deduplicating by name would *hide* the real source behind the broken row — the user would be left with an entry that does nothing and no way to add the one that works. That's a worse outcome than the duplicate.

There is also no way to tell a legacy row from a row whose extension was simply uninstalled. Both are just a `Long` that resolves to nothing. Any migration would have to guess between them.

**What it does instead.** A row whose key no loaded source owns now renders as **"Source not installed"** with its toggle disabled, and the real source stays addable. That needs no guessing, deletes nothing, and gives the same correct affordance for both cases — delete it, or install the extension. The one action that makes sense either way is the one the UI now points at.

Two tests, mutation-checked: hardcoding `isInstalled = true` fails the one that names the behaviour.

Worth noting this is the third time in two PRs that the honest answer to "these two keys disagree" has been *surface it* rather than *reconcile it* — same as `resolveDownloadFolderName` (#1256) and the legacy-key fallback in `getSourceByKey`. When a `Long` on disk carries no record of which rule wrote it, a migration cannot separate the cases, so the choice is between guessing and admitting.

---

